PR #52, opened 2020-10-25
Make javax.annotation-api dependency optional

norrisjeremy:
Use reflection to find javax.annotation.Priority, thus making the javax.annotation-api dependency optional.

### DIFF
--- a/inject/pom.xml
+++ b/inject/pom.xml
@@ -30,12 +30,6 @@
     </dependency>
 
     <dependency>
-      <groupId>javax.annotation</groupId>
-      <artifactId>javax.annotation-api</artifactId>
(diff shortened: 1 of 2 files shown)


rbygrave:
Yes, I like this change.  I think we can improve it a little by moving the `Class.forName("javax.annotation.Priority")` earlier so I'll look to do that adjustment after merging this.

norrisjeremy:
Excellent, thanks! We make use of Guice in many internal projects and I'm interested in moving over to this as an alternative so that the injection is all handled at compile time. It's just that we don't make use of javax.annotation-api, so I was looking at how it could be left out.

Issue #51, opened 2020-09-07
Detect circular dependency and fail compile with nice error - Was BeanContextFactory.createContext calls builders in the wrong order


aletundo:
Hi @rbygrave,

It's not the first time I'm facing this issue. Even if `@DependecyMeta` annotation contains the correct classes within the `dependsOn` list, the `createContext` method calls the builders in the wrong order causing:

`Injecting null for com.example.MyClass when creating class com.example.MyClass2 - potential beans to inject: []`

The generated `createContext` methods is like:

```java
  @Override
  public BeanContext createContext(Builder parent) {
    builder.setParent(parent);
    // other builders
    build_MyClass2();
    // other builders
    build_MyClass();
    return builder.build();
  }
```

Any workarounds I can implement to force the generator to call the builders respecting the dependencies?

Thanks a lot 🙏

rbygrave:
Are you able to provide a test case?  

The build order should honour/respect the dependencies so we need to see why that isn't the case.

aletundo:
Hi,

I shared [here](https://gitlab.com/learnERC/varys/-/tree/master/varys-monitoring-claim-controller/src/main/java/io/varys/monitoringclaimcontroller) the source code of the affected project.

Since one of the dependencies is not yet avaialable on a Maven public repository, I'm attaching the generated `_di$Factory.java` source. Otherwise, `varys-api-jclient` project (in the same repo) needs to be installed with Maven to success in compiling the affected project.

```java
// package ...
// imports ...
@Generated("io.avaje.inject.generator")
@ContextModule(name="io.varys.monitoringclaimcontroller")
public class _di$Factory implements BeanContextFactory {

  private final Builder builder;

  public _di$Factory() {
    this.builder = BuilderFactory.newBuilder("io.varys.monitoringclaimcontroller", null, null);
  }

  @Override
  public String getName() {
    return builder.getName();
  }

  @Override
  public String[] getProvides() {
    return builder.getProvides();
  }

  @Override
  public String[] getDependsOn() {
    return builder.getDependsOn();
  }

  @Override
  public BeanContext createContext(Builder parent) {
    builder.setParent(parent);
    build_ObjectMapperFactory();
    build_ApiFactory();
    build_RedisConfig();
    build_ChmmServiceImpl();
    build_ObjectMapper();
    build_ApiClient();
    build_MonitoringClaimApi();
    build_MonitoringUnitApi();
    build_ProbeApi();
    build_RedisClientFactory();
[...]

PR #59, opened 2020-10-29
#51 Add jpms module-info using Multi-Release jar

rbygrave:


inject/README.md
```diff
@@ -1,6 +1,20 @@
...
+# Avaje Inject
 APT based dependency injection for server side developers - https://dinject.io
 
+### Example module use
+
+```java
+module org.example {
+
+  requires javax.inject;
+  requires org.slf4j;
```
norrisjeremy, line 10: Do consumers actually need to add SL4J?

inject/src/main/java9/module-info.java
```diff
@@ -0,0 +1,12 @@
+module io.avaje.inject {
+
+  exports io.avaje.inject;
+  exports io.avaje.inject.core;
+
+  requires javax.inject;
+  requires org.slf4j;
+  requires static org.mockito;
```
norrisjeremy, line 8: Should there be a `requires static javax.annotation` so that the optional reflection for `@Priority` works?

inject-generator/src/main/java9/module-info.java
```diff
@@ -0,0 +1,8 @@
+module io.avaje.inject.generator {
+
+  requires java.compiler;
+  requires java.annotation;
+  requires io.avaje.inject;
```
norrisjeremy, line 5: Should there be a `requires javax.inject`?

inject/README.md
```diff
@@ -1,6 +1,20 @@
...
+  requires transitive io.avaje.inject;
```
norrisjeremy, line 11: Just curious as to why it has to be transitive?

inject/README.md
rbygrave: No, I got that wrong ... because in the inject module-info I needed to make these two dependencies `requires transitive`.  Once I do that the consumers no longer need to specify `requires org.slf4j` or  `requires javax.inject;`.

inject/src/main/java9/module-info.java
rbygrave: I think so - need to check that.  Also note I need to add in `transitive` for javax.inject and org.slf4j.

inject/README.md
rbygrave: Nice catch, it doesn't have to be.  Fixing and updating this.

inject-generator/src/main/java9/module-info.java
rbygrave: I'm going to say **_Yes_** (and add it).  It has been working for me without a `requires javax.inject` but that doesn't seem right - perhaps related to having an unnamed module (java.annotation) in there.  Hmmm.

Adding javax.inject as that seems more correct.  In theory javax.inject comes in with io.avaje.inject now but I've only just made that change and it was working before so confused as to why it was working.

norrisjeremy:
Have you given any thought to migrating to new jakarta.inject / jakarta.annotation-api? It might be good time to do the migration while you are introducing JPMS.

inject/src/main/java9/module-info.java
norrisjeremy: I'm not sure if org.slf4j needs to be transitive? It's my understanding that transitive is only required if types from the dependency are exposed in your exported packages. So if I understand correctly, unless one of the classes from the exported packages returns a type from org.slf4j or consumes a type from org.slf4j (like as an argument to a method), then I don't believe it needs to be declared transitive.
But I could be totally misunderstanding (I've not done alot of JPMS work yet).

inject-generator/src/main/java9/module-info.java
norrisjeremy: I think if you decide to declare javax.inject as transitive in io.avaje.inject, then it wouldn't be necessary to then require it here (since it would then be picked up from io.avaje.inject).

inject/src/main/java9/module-info.java
rbygrave: So it works without the `requires static java.annotation`.   I _believe_ that is because java.annotation is an unnamed module - so all packages in java.annotation are treated as open allowing reflection on all public classes.

rbygrave:
> Have you given any thought to migrating to new jakarta.inject / jakarta.annotation-api? 

I did a little bit. jakarta.inject uses Automatic-Module-Name in the manifest.

> It might be good time to do the migration while you are introducing JPMS.

Well I'm still learning but where I'm at at the moment is that I think it is ok for some modules to be "unnamed" ones.  My working theory is that unnamed modules are ok in the case that:
- It has no dependencies 
- If it had a module-info it would just contain ... export * / export all packages.

As I see it right now, both javax.inject and java.annotation-api match this description (and my example jpms app is fine with them as they are).  So right now if I was doing a full jpms app those modules are actually fine to use as is. I currently don't see anything to gain from say forking and making a jpms version of them (some people have done this) or moving to say jakarta.inject (even if they had a module-info which they don't).

Hmmm.

inject/src/main/java9/module-info.java
rbygrave: > if org.slf4j needs to be transitive

If it isn't transitive the consumer module needs to declare it in their module-info as `requires org.slf4j`.  Without that code declaring org.slf4j.Logger does not compile.  So I'm happy that it should be transitive in this case.

inject-generator/src/main/java9/module-info.java
rbygrave: Yes.  I added it thinking it wouldn't harm being explicit but now thinking I should really leave it up to io.avaje.inject to bring it in (and remove it again).

inject/src/main/java9/module-info.java
norrisjeremy: > If it isn't transitive the consumer module needs to declare it in their module-info as `requires org.slf4j`. Without that code declaring org.slf4j.Logger does not compile. So I'm happy that it should be transitive in this case.

Ok, that seems to make sense.

norrisjeremy:
> Well I'm still learning but where I'm at at the moment is that I think it is ok for some modules to be "unnamed" ones. My working theory is that unnamed modules are ok in the case that:
> 
> * It has no dependencies
> * If it had a module-info it would just contain ... export * / export all packages.
> 
> As I see it right now, both javax.inject and java.annotation-api match this description (and my example jpms app is fine with them as they are). So right now if I was doing a full jpms app those modules are actually fine to use as is. I currently don't see anything to gain from say forking and making a jpms version of them (some people have done this) or moving to say jakarta.inject (even if they had a module-info which they don't).
> 
> Hmmm.

Part of my thought process for suggesting a migration to the new 2.x jakarta.inject + jakarta.annotation-api versions are two-fold:

1. I believe the module names for the old javax versions are technically reserved by the JDK itself (javax.inject / java.annotation). While I would think it's unlikely Oracle would decide to add modules with colliding names into the JDK itself, it's possible they could. And then I think hilarity could ensue, since there would be conflicting module names.

2. With adding in the JPMS support, it might be best to rip the bandaid off entirely all at once, and shift everything to using the new jakarta.inject + jakarta.annotation package names all at the same time. This would be admittedly highly disruptive to users of this library, since they would need to adjust all their classes to use the jakarta.inject + jakarta.annotation annotations instead of javax.inject + javax.annotation.

Anyway, that was just part of my thinking at the moment (while reviewing our internal projects and determining how best to migrate them from Guice to this library).

rbygrave:
> it might be best to rip the bandaid off entirely all at once, and shift everything to using the new jakarta.inject + jakarta.annotation package names all at the same time. 

Maybe, although with this change it still supports java 8 and classpath.  This change doesn't demand anything from it's users per say. It's ready for java modules but the majority are still going to be classpath for a while I'm suspecting.  


> This would be admittedly highly disruptive to users of this library,

Well it would. 

What I have in my mind is that there might literally be 2 changes to go in with:
A) this change 
B) maybe rename core package to spi.  (pondering this)

Once those are in this library could be very stable and literally only need bug fixes. So my thinking at the moment is to get to there ... be nice and stable and boring ... wait for demand to go to jakarta.inject.  

Ideally the only change after A) and B) is ... user demand to migrate to jakarta.inject (from javax.inject).

norrisjeremy:
> Well it would.
> 
> What I have in my mind is that there might literally be 2 changes to go in with:
> A) this change
> B) maybe rename core package to spi. (pondering this)
> 
> Once those are in this library could be very stable and literally only need bug fixes. So my thinking at the moment is to get to there ... be nice and stable and boring ... wait for demand to go to jakarta.inject.
> 
> Ideally the only change after A) and B) is ... user demand to migrate to jakarta.inject (from javax.inject).

Yep, that probably makes the most sense. I tend to think too big sometimes :).

rbygrave:
Although maybe we could/should have both by maintain 2 branches - a branch for javax.inject based dependencies and a branch for jakarta.inject dependencies.  

My situation sounds similar with existing Guice, HK2 (Jersey) and Spring - and the migration is not always full replacement [of Guice] but where there is co-existence moving towards replacement over time.  Now with co-existence I think it would have been nicer/easier to use `jakarta.inject` so that there was clear distinction.

**Plan**
- Release what's in master as 1.4
- Merge in the jpms change, refactor rename core to spi - release as 2.0
- Create a long lived branch called "jakarta" and in that branch move to jakarta dependencies, release off that branch as 3.0

Developers can then choose version 2.0 (javax.inject) or 3.0 (jakarta.inject)
Bug fixes going forward need to apply to both branches and release 2.x and 3.x when that occurs.

norrisjeremy:
> * Release what's in master as 1.4
> * Merge in the jpms change, refactor rename core to spi - release as 2.0
> * Create a long lived branch called "jakarta" and in that branch move to jakarta dependencies, release off that branch as 3.0
> 
> Developers can then choose version 2.0 (javax.inject) or 3.0 (jakarta.inject)
> Bug fixes going forward need to apply to both branches and release 2.x and 3.x when that occurs.

That sounds like it could be workable plan. It probably wouldn't be too difficult to keep any changes in sync between 2.x branch & 3.x branch, since I think the main difference would just be a search/replace of javax.inject => jakarta.inject & javax.annotation => jakarta.annotation.

PR #66, opened 2020-10-30
Add support for method based injection.

norrisjeremy:
This is a missing feature I stumbled across when attempting to convert a project from Guice, which supports `@Inject` on methods.
I tweaked a few classes in the inject-test module to use method injection instead of field injection and the unit tests all still passed.
I also examined the generated code and it seemed correct to me, so hopefully I didn't miss anything.


rbygrave:
Cool.  Looks good.

PR #71, opened 2020-11-02
Add io.avaje.inject @PostConstruct/@PreDestroy annotations

rbygrave:
We can use any `@PostConstruct` or `@PreDestory` annotation.  Adding these as avaje inject annotations means that people don't need to depend on the JEE javax.annotation or jakarta.annotation for these annotations.

That is, we consider them important enough to have our own (and dependency on jakarta.annotation for these is not ideal).


norrisjeremy:
What do you think about adding an `@Priority` as well?

rbygrave:
> What do you think about adding an @Priority as well?

I think it's a good idea.  I almost have a PR that take any Priority annotation.

PR #72, opened 2020-11-02
Support using any Priority annotation for sorting

rbygrave:


rbygrave:
FYI @norrisjeremy - I think adding a `@Priority` would be good and then probably it should be the default one used rather than the (old JEE) javax.annotation.Priority one.

norrisjeremy:
> FYI @norrisjeremy - I think adding a `@Priority` would be good and then probably it should be the default one used rather than the (old JEE) javax.annotation.Priority one.

Yes, I would concur. Anything to avoid pulling in the old JEE annotation-api is a positive step.

PR #73, opened 2020-11-02
Add @Priority and make it the default

rbygrave:


norrisjeremy:
I like this, as it allows using Priority based injection without having to pull in potentially problematic javax.annotation-api dependency.

inject-test/src/test/java/org/example/coffee/priority/base/CBasei.java
```diff
@@ -0,0 +1,15 @@
...
+
+import javax.inject.Singleton;
+
+@Singleton
+@Priority(1)
+public class CBasei implements BaseIface {
+
+  @Override
+  public String other() {
+    return "a";
```
norrisjeremy, line 13: Minor nit, but should this be `return "c";`?

inject-test/src/test/java/io/avaje/inject/SystemContextTest.java
```diff
@@ -16,28 +17,30 @@
   @Test
   public void getBeansByPriority() {
 
-    final List<Somei> beans = SystemContext.getBeansByPriority(Somei.class);
-    assertThat(beans).hasSize(2);
+    final List<Somei> beans = SystemContext.context().getBeansByPriority(Somei.class, Priority.class);
```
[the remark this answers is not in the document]
rbygrave: Yes, I'll merge and sort that out - thanks.

inject-test/src/test/java/org/example/coffee/priority/base/CBasei.java
rbygrave: Yes it should, it will be nice to fix it - thanks

PR #75, opened 2020-11-02
Tweaks for PostConstruct/PreDestory changes

norrisjeremy:
1. Allow PostConstruct/PreDestory annotations from other packages.
2. Remove javax.annotation-api dependency since it is no longer strictly required.

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
@@ -264,14 +264,12 @@ class BeanReader {
       }
     }
 
-    if (context.isPostConstructAvailable()) {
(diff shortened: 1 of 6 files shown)


inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
```diff
@@ -264,14 +264,12 @@ private void readMethod(Element element, boolean factory) {
       }
     }
 
-    if (context.isPostConstructAvailable()) {
```
rbygrave, line 267: Thanks for spotting that ... ah, tests passed because it was still in the classpath - nice !!

rbygrave:
Excellent, thanks for that.

rbygrave (approved):

PR #76, opened 2020-11-03
Add javadoc comments into generated code

rbygrave:


inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
```diff
@@ -53,7 +53,7 @@
   }
 
   private String generatedAnnotation(boolean jdk8) {
-    return jdk8 ? null : isTypeAvailable(Constants.GENERATED_9) ? Constants.GENERATED_9 : null;
+    return jdk8 ? Constants.GENERATED_LOCAL : isTypeAvailable(Constants.GENERATED_9) ? Constants.GENERATED_9 : Constants.GENERATED_LOCAL;
```
norrisjeremy, line 56: What about just switching to always using a local `@Generated` annotation regardless of the JDK version or if JPMS is in use?
I think that's what Lombok finally decided to do.

norrisjeremy:
I like this too!

inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
rbygrave: Yes, that is pretty tempting. I've had a little discussion on the jigsaw mailing list about this and at this stage I'm pretty confident that we ultimately can't use `javax.annotation.processing.Generated` with JPMS so just using the local one makes more and more sense to me now.

PR #78, opened 2020-11-03
Simplify to just use local @Generated annotation

rbygrave:


inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
```diff
@@ -22,24 +22,16 @@
```
norrisjeremy, line 22: I think the `javax.lang.model.SourceVersion` import is unused now?

Yedfree (approved):

PR #134, opened 2021-07-19
Refactor scopes - Removes ApplicationScope & RequestScope in favour of "custom scopes"

rbygrave:
- Removes ApplicationScope  - Migrate to BeanScope.newBuilder().build()
- Removes RequestScope - Migrate to using custom scopes
- Removes deprecated methods
- Moves `withSpy()` `withMock()` methods of `BeanScopeBuilder` to only be available after `beanScopeBuilder.forTesting()` 
- Changes `@InjectModule` to use `Class<?>` for provides and requires (not strings)
- Changes BeanScopeFactory to instead be `Module` and allow us to specify explicitly which modules to include in BeanScope
- Adds support for "custom scopes", each scope generates it's own `Module`
- Allows BeanScope to have a parent BeanScope


mikehearn:
Nice, from a quick scan through it feels like much more code was deleted than added, but without loss of expressive poewr.

rbygrave:
Yes, so far this is has the "good feeling" about it.  I have just managed some initial "custom scope" support.  I think I have about 3 more things to support for custom scopes.

--- 
Splits generation of old BeanScopeFactory into Module + a helper BeanFactory. This allows the use of the generated Module via BeanScopeBuilder.withModules().

**This does not support**:
- partial compile (need to reload meta data for custom scope modules on partial compile)
- external/requires dependencies provided by other scopes / typically the main scope
- parent/child relationship of BeanScopes (e.g. custom scope with parent of the global scope)

rbygrave:
### Next steps:

- parent/child relationship of BeanScopes

At this point we allow a custom scope to depend on anything provided by the "default" scope - effectively assuming that the custom scope will be created with the "default" scope as it's parent.  At least, this is the initial plan as that allow us to get an initial release out for this.

Just need the parent / child support and then we can take stock and look / play with it.

rbygrave:
@mikehearn   I have released this as version 6.5-RC0 to maven central. I don't have anything outstanding to do on it, it is ready to try out.

rbygrave:
WRT the prior code example:
```kotlin

@CreateScope
annotation class MyGlobal

@CreateScope(external = [Foo::class, Bar::class], parent = MyGlobal::class)
annotation class MyRequest

@MyGlobal class A
@MyRequest class B(val a: A)

fun main() {
  val scope = MyGlobalScope.newBuilder().build()
  val requestScope = MyRequestScope.newFrom(scope).build()
}

```
### Default scope

There is still this concept of the default scope (global scope). What this boils down to is that when we are not using custom scopes those beans/components goto a 'default' Module. When we build a BeanScope without specifying any explicit modules what happens is that is will use ServiceLoader to find and load all the 'default' Modules in the classpath/modulepath.

So we can use avaje-inject in many jars each with a 'default' Module, when we create BeanScope without specifying any explicit modules it loads all of these and wires the whole lot. So `BeanScope.newBuilder().build()` does this.

Now this is generally what we do when we build reusuable components or modular monoliths - we are wiring things from multiple jars into a single BeanScope.

Note that with 6.5.RC0 we don't need to use this 'default' scope.  We can do everything as custom scopes.


### Custom scope

Pretty similar to your original example
```java
package org.example;

import io.avaje.inject.InjectModule;
import jakarta.inject.Scope;
import org.example.custom.LocalExternal;

@Scope
@InjectModule(requires = {System.class, LocalExternal.class})
public @interface MyCustomScope {
}MyGlobal

``` 
```java
package org.example;

import io.avaje.inject.InjectModule;
import jakarta.inject.Scope;
import org.example.custom.LocalExternal;

@Scope
@InjectModule(requires = {System.class, LocalExternal.class})
public @interface MyCustomScope {
}MyGlobal

```

rbygrave:
Notes:

The generated 'default' module implements `io.avaje.inject.spi.Module` and has a `META-INF/services/io.avaje.inject.spi.Module` file to match (for service loading all the 'default' modules).

The generated 'custom' modules implement `io.avaje.inject.spi.Module.Custom` ... and have a `META-INF/services/io.avaje.inject.spi.Module.Custom` file to match but avaje-inject is not service loading the custom modules. In theory we don't need the services file for these custom scopes but they could be used if someone wanted to discover all the custom scopes in the classpath (I don't know why someone would want to do this yet, maybe it is something no one ever does).

Issue #140, opened 2021-08-02
Are scope requirements meant to be transitive?


mikehearn:
I wrote the moral equivalent of this:

```

@Scope
@InjectModule(requires = [ A::class, B::class ])
annotation class Foo

@Scope
@InjectModule(requires = [ Foo::class, C::class ])
annotation class Bar

@Bar
class Baz(val a: A, val b: B)
```

I was expecting that because scope Bar depends on scope Foo, and Foo depends on A (i.e. it's externally provided) that Baz would be able to depend on A. However it seems that doesn't work. I get errors about missing dependencies. Is that intentional? If so, is the right thing to do just duplicating the requires=[] list ?

rbygrave:
> Baz would be able to depend on A

Yes, that should work. Baz should be able to depend on A.

> However it seems that doesn't work. I get errors about missing dependencies. 

Should be a bug. 

We have `ParentScopeTest` but that doesn't test external dependencies on the parent scope which is the case here with the Foo scope above.  I'll look to create a test case where the parent scope has external dependencies and get that to reproduce.

rbygrave:
Ok, fixed in 6.6

mikehearn:
Awesome, thanks.

PR #141, opened 2021-08-03
 #140 - Are scope requirements meant to be transitive?

rbygrave:
If a scope depends on a parent scope and that parent scope has
external dependencies - then the scope should also be able to
depend on those external dependencies.

Ultimately the fix for this is in ScopeInfo.providesDependency()
adding in the check to see if the dependency is in requires
(is an external dependency of the scope).


rbygrave:
This was merged. Closing.

Issue #156, opened 2021-10-06
Dependency on SLF4J - Migrate to System.Logger


XakepSDK:
Can dependency on SLF4J be removed or library updated to 2.0-alpha? 1.7 is not modularized.

rbygrave:
I have had a closer look at slf4j 2.x to see what was going on and why it's still alpha etc.  Also just a little look at `java.lang.System.Logger`. For myself, I've been for so long just using slfj4-api that it had become a strong bias and I kind of missed `java.lang.System.Logger` arriving and yes it does look good to me. 

That is, I'm starting to remove my bias away from slf4j-api and leaning towards the thought that `java.lang.System.Logger` might be the longer term better option when compared to slf4j-api 2.x. Given that avaje-inject has such little need for logging anyway it makes it more obvious for avaje-inject that we should probably push it towards the standard java logger(s)   

> Use multi-release jar to use SLF4J 1.7 for Java 8 and Platform Logger for Java 9+ 

We could do this or even use JUL for Java 8 and Platform Logger for 9+.  At some point we also raise the question of moving our minimal java dependency from 8 to 11 which would mean we should just use the Java Platform Logger / `java.lang.System.Logger` and wouldn't need multi-release jar (but MR-jar should also be a pretty good and simple approach here).


**Next steps**
I think the next step is to try and prepare a PR that looks to take a multi-release jar approach with JUL and Platform Logger and see what that looks like.

rbygrave:
Note that I have the PR/branch https://github.com/avaje/avaje-inject/pull/172/files ... but I am somewhat unsatisfied with it at the moment.

My thought is that we almost certainly should change to `System.Logger`.  If we bump to Java 11 then this is pretty simple and I am thinking that is maybe the best long term path going forward.  To bump to Java 11 and at that time also change to use `System.Logger`.

https://www.reddit.com/r/java/comments/rdv98z/have_you_ever_wondered_how_javas_logging/ho49gp1/

rbygrave:
PR #190 bumps to Java 11 and uses System.Logger.  Closing this as now fixed.

PR #172, opened 2021-12-14
#156 - Multi-release jar with logging via System.Logger and JUL

rbygrave:
This isn't fully satisfactory yet.


rbygrave:
Not going to merge this, instead just going to use System.Logger with a bump to Java 11.

PR #181, opened 2021-12-22
Fixed implementation for previous PR

doppelrittberger:
Fixed implementation for bugfix. I missed to re-check. Sorry :)

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
@@ -9,6 +9,8 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 class MethodReader {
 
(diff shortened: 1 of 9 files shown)


inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
```diff
@@ -179,7 +181,21 @@ void addImports(Set<String> importTypes) {
...
+  void factoryImports(Set<String> importTypes) {
+    Set<GenericType> genericTypes = getGenericTypes();
+    if (!genericTypes.isEmpty()) {
+      importTypes.add(Constants.TYPE);
+      importTypes.add(Constants.GENERICTYPE);
+      importTypes.addAll(genericTypes
+        .stream()
+        .flatMap(g-> Stream.concat(
+          Stream.of(g.getMainType()),
+          g.getParams().stream().map(GenericType::getMainType)))
```
rbygrave, line 196: Just noting that GenericType.addImports() does this for us.  I added a commit to do this part so I think I'll merge this in and then trim out the bits that are now duplicated.

Good stuff !!!

doppelrittberger: Sounds good 👍

Issue #180, opened 2021-12-22
Static inner classes as dependencies


doppelrittberger:
Hi,

if I use static inner classes as bean definitions Avaje seems to struggle. It creates a package which is named like the parent class and this package clashes with the parent class itself.
```
public class MyBeans {
@Singleton
public static class Bean1 {}
}
```
Above example generates a class in package ``MyBeans`` and this name clashes with the name of the class ``MyBeans``.

If a PR is ok please let me know.

Thanks
Markus

rbygrave:
Yes, a PR would be great !!

rbygrave:
Fix PR merged into master, closing.

PR #182, opened 2021-12-23
added fix and test for #180

doppelrittberger:
Hi,

this is the PR for https://github.com/avaje/avaje-inject/issues/180.

Thanks


rbygrave:
Ok, I've merge this into master branch. It isn't showing it's merged because the PR is to merge into `avaje:doppelrittberger-factory-types-bugfix` and I didn't notice until after I merged it.  I **_should_** have changed that BEFORE I merged it in darn it.

Anyway, this is merged in but just not showing it so I'll just hit the close button instead.

PR #252, opened 2023-01-05
 #251 - Adjustments for fixing Aspects on methods with generic parameters 

rbygrave:


SentryMan (approved):
LGTM, yeah putting a full generic in constructor is good.

Issue #258, opened 2023-01-12
ENH: Add Plugin API for 'default providers' and BeanScopeBuilder.provideDefault() methods


rbygrave:
This allows plugins to add a 'default provider' (as a `@Secondary`) that can be used when nothing else provides that dependency.

For example, Jsonb can have a plugin to use `() -> Jsonb.builder().build()` as a provider like:

```java
/**
 * Plugin for avaje inject that provides a default Jsonb instance.
 */
public final class DefaultJsonbProvider implements io.avaje.inject.spi.Plugin {

  @Override
  public Class<?>[] provides() {
    return new Class<?>[]{Jsonb.class};
  }

  @Override
  public void apply(BeanScopeBuilder builder) {
    builder.provideDefault(Jsonb.class, () -> Jsonb.builder().build());
  }
}
```

SentryMan:
So if a dependency provides a plugin, I don't need to define @InjectModules explicitly? Or did I misread this entirely?

rbygrave:
Yes, you got it.

Jsonb and the Validator are 2 good examples. For something to be a plugin:

1. It only provides one component
2. The component it provides must not have any dependency
3. The component it provides is not named/qualified and has `@Secondary` priority so will be used only when nothing else provides the same component

So we'd use a Plugin rather than a full blown avaje-inject module with `@InjectModules` etc because with the plugin it ends up as approximately 2 lines of code and a full blown avaje-inject module instead looks a bit more like serious overkill.

SentryMan:
Yeah I figured. I'm getting a dependency not provided error on my javalin repo when I use the new validator. I guess the service loader can't load the plugins at compile time. #263 fixes this.

PR #259, opened 2023-01-12
#258 - ENH: Add Plugin API for 'default providers' and BeanScopeBuilder.provideDefault() methods

rbygrave:


SentryMan:
Inspiration is really striking, isn't it? I like it.

PR #264, opened 2023-01-23
Now Aspect Targets are Added to AutoProvides

SentryMan:
Now modules can auto-expose their aspect implementations for wiring.

- add method to ProcessingContext that can find the target AspectProvider class name
- Update SimpleModuleWriter to directly add aspect targets to autoProvides
- make @Aspect only accept classes that implement AspectProvider.
- remove buildAutoProvidesAspects method
- remove autoProvidesAspects

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
@@ -1,11 +1,24 @@
 package io.avaje.inject.generator;
 
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.io.LineNumberReader;
+import java.io.Reader;
+import java.nio.file.NoSuchFileException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
 import javax.annotation.processing.Filer;
 import javax.annotation.processing.FilerException;
 import javax.annotation.processing.Messager;
 import javax.annotation.processing.ProcessingEnvironment;
 import javax.lang.model.element.Element;
 import javax.lang.model.element.TypeElement;
+import javax.lang.model.type.MirroredTypeException;
 import javax.lang.model.type.TypeMirror;
 import javax.lang.model.util.Elements;
 import javax.lang.model.util.Types;
@@ -13,12 +26,8 @@ import javax.tools.Diagnostic;
 import javax.tools.FileObject;
 import javax.tools.JavaFileObject;
 import javax.tools.StandardLocation;
-import java.io.FileNotFoundException;
-import java.io.IOException;
-import java.io.LineNumberReader;
-import java.io.Reader;
-import java.nio.file.NoSuchFileException;
-import java.util.*;
+
+import io.avaje.inject.aop.Aspect;
 
 final class ProcessingContext {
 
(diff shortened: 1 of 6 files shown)


rbygrave:
> Now modules can auto-expose their aspect implementations for wiring.

Why do this?  What was the need driving this change?

Aspect target / **implementations** are NOT expected to have any component in _another module_ depend on them. Other modules should only need/use the aspect (annotation) and should not have any dependency on the aspect implementation.

That is, I am not expecting to see the aspect _implementation_ in `autoProvides()` ... but I do expect to see the aspect _annotation_ in `autoProvides()` and this removes that.

SentryMan:
Basically, I wanted to get aspects to work automatically without having to add @InjectModules(requires=aspecttarget.class)

rbygrave:
Yes but I don't understand fully and I don't think this "fixes" the issue per say. 

We should not say `requires=aspecttarget.class` ... but instead `requires=aspectAnnotation.class`.  Our module using the aspect should really only know about the annotation.

> without having to add @InjectModules(requires=aspecttarget.class)

How did this change fix that?

rbygrave:
The "Fix" or expectation in my mind is that `autoRequires()` includes the _aspect annotations_ that are being used on the components being wired by the module.

SentryMan:
> How did this change fix that?

Now we no longer need the @InjectModules to get an aspect working. Basically, I wanted to have a common aspect library that "just works" without a need to explicitly add provides and requires.

rbygrave:
> hat "just works" without a need to explicitly add provides and requires.

I don't see how this PR makes that "just work" without a change/fix to the "requires" side. That is, if there is an issue here it is on the "requires" side.  

This PR changes `autoProvides` to not include the annotation but include the implementation. For that to "fix" a problem suggests to me that meant there was a module that had a dependency on an aspect implementation?  That doesn't make sense to me - can you explain that?

rbygrave:
> I wanted to have a common aspect library that "just works" without a need to explicitly add provides and requires.

This should actually work today with the released 8.12-RC2 in maven central without this change. So I'm trying to understand why that does not work and why this PR seems to fix the issue.

SentryMan:
Say I got this aspect `@Timed` in module A, and I try to use @Timed on a method in module B. The generated proxy class in module b is:
```
@Proxy
public final class ServiceClass$Proxy extends ServiceClass {

  private final TimedAspect timedAspect;

  private Method callDownStream0;
  private MethodInterceptor callDownStream0Timed;

  public ServiceClass$Proxy(TimedAspect timedAspect, ApiClient api) {
    super(api);
    this.timedAspect = timedAspect;
    try {
      callDownStream0 = ServiceClass.class.getDeclaredMethod("callDownStream");
      callDownStream0Timed = timedAspect.interceptor(callDownStream0, callDownStream0.getAnnotation(Timed.class));

    } catch (Exception e) {
      throw new IllegalStateException(e);
    }
  }

  @Override
  public byte[] callDownStream() {
    var call = new Invocation.Call<>(() -> super.callDownStream())
      .with(this, callDownStream0);
    try {
      callDownStream0Timed.invoke(call);
      return call.finalResult();
    } catch (InvocationException e) {
      throw e;
    } catch (Throwable e) {
    }
  }
}
```
Currently, module b fails to compile with 
```
[ERROR] /M:/Dev/avaje-shaded-module-example/target/generated-sources/annotations/com/jojo/javalin/shade/service/ServiceClass$Proxy.java:[17,14] No dependency provided for com.jojo.javalin.api.aspect.TimedAspect on com.jojo.javalin.shade.service.ServiceClass$Proxy
```
It requires the TimedAspect in the dependency.

rbygrave:
> No dependency provided for com.jojo.javalin.api.aspect.TimedAspect

The current generated proxy code is depending on `com.jojo.javalin.api.aspect.TimedAspect` (the concrete aspect target implementation) when it _could_ instead depend on `io.avaje.inject.aop.AspectProvider<Timed>`. Which would be that one level of indirection away from the concrete implementation of the aspect.

Right now that pretty much requires the aspect implementation to be public. Using `AspectProvider<Timed>` also could have improvements in terms of using test doubles for the aspects.

So to me the first issue is to ponder if the generated proxy code should instead use `AspectProvider<Timed>` ... in which case we've moved the issue / problem a little bit.

SentryMan:
I guess this PR isn't the way to do it

PR #265, opened 2023-01-25
Refactor AutoProvides

SentryMan:
Now we can control what beans get auto-provided through `@InjectModules`.

- adds new parameter to `@InjectModules` to control auto provide strategy
- adds  strategy to get all top level interfaces/classes
- adds  strategy to get all beans as they are defined (Was talking to a dev that was trying to autoprovide a customized ObjectMapper, but got the technically true but useless `Versioned` interface )
- adds strategy to not autoprovide at all.

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
@@ -25,7 +25,7 @@ import javax.tools.FileObject;
 import javax.tools.JavaFileObject;
 import javax.tools.StandardLocation;
 
-import io.avaje.inject.InjectModule.AutoProvideLevel;
+import io.avaje.inject.InjectModule.AutoProvideStrategy;
 
 final class ProcessingContext {
 
@@ -36,7 +36,7 @@ final class ProcessingContext {
   private final Types typeUtils;
   private final Set<String> uniqueModuleNames = new HashSet<>();
   private final ExternalProvider externalProvide = new ExternalProvider();
-  private AutoProvideLevel autoProvideLv;
+  private AutoProvideStrategy autoProvideStrategy;
 
   ProcessingContext(ProcessingEnvironment processingEnv) {
     this.processingEnv = processingEnv;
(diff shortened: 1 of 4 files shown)


SentryMan:
I assume then that I should take out that else clause and only auto provide the top level superclasses/interfaces?

rbygrave:
>  only auto provide the top level superclasses/interfaces?

The original thinking is that for inter-module dependencies these should really only be interfaces - purely abstract.

Hence that is what it does currently - ONLY interfaces get into `autoProvides()`. 

So my question is, what is the motivation here that made you want to make this change? Did you have an inter-module dependency that was not an interface?

SentryMan:
> So my question is, what is the motivation here that made you want to make this change? Did you have an inter-module 

Well not me specifically, another developer asked me why only interfaces worked. He had an abstract class in a common mongo lib he wanted to autoprovide.

I added that else clause because I thought it would be nice for utility libraries made with avaje to "just work" without going through the (admittedly minor) hassle of explicitly defining provides and requires. 

> The original thinking is that for inter-module dependencies these should really only be interfaces - purely abstract.

also this was unknown to me.

SentryMan:
How about we call it a draw and just allow top-level classes?

rbygrave:
> How about we call it a draw and just add on top-level classes?

Ha ha.

Well I've done some pondering and actually I think I need to concede a bit more than that. 

**_TLDR_** we should probably merge this (or something as close as damn it to this).

The "more purist" design of "interfaces only" doesn't handle the cases of modules we create for "Utility" or "Plumbing" purposes (like the Mongo case and others I think we would similarly hit in the case of some messaging libraries. In these cases we developers don't provide the types of the components we looking to provide but instead they are provided by the 3rd party library being used and the library provides abstract classes rather than interfaces).

Then it boils down to the question of what the consequences are of "provides" having more types listed (maybe a lot more) - and I think the answer is that it doesn't practically matter too much but it might be a little harder to read with bigger modules (lots of types in "provides").

Note: A likely future extension / bugfix is that with a `module-info` we probably should read and honor the `exports` packages. Filter the "autoProvides" to only include types in the exports packages.

SentryMan:
@rbygrave what do you think about this? I've made it so we can control through @InjectModules.

rbygrave:
It might be a good idea, I'm just not sure of that yet.

WRT ALL ... and https://github.com/avaje/avaje-inject/pull/265/files#diff-218c0443b7520974d092b9294faeb9167b3e93e3d26da339215c5dc2c286e6c3R116-R118 ... the treatment of aspect provides has changed in the code I'm about to merge so that ALL part isn't quite right. In my mind if we were to to "ALL" then we need to change it to return `List<String>`.

So right now I'm not sure about the ALL and the NONE.  I do think we need to include top level classes (not just interfaces). So I think I'll manually incorporate those changes into the branch I've got.
- Include stopping at `java.lang.Record`
- autoProvides includes top level non-interfaces types

SentryMan:
I got what I initially came here for.

The only "issue" with interface type as they are now is that when you occasionally have to deal with 3rd party classes that extend a whole chain of interfaces. (Like that ObjectMapper case)  In this case, we'd need to add `provides=ObjectMapper.class` to the module. I can deal with that. Will reopen if enough people ask.

PR #270, opened 2023-01-31
Update Aspect generation - add aspectProvides() meta data

rbygrave:


rbygrave:
Note these commits have been included in https://github.com/avaje/avaje-inject/pull/270

PR #271, opened 2023-01-31
Add Aspect Auto Requires

### LINKED ISSUE #269: AutoProvideAspects doesn't work as expected
Say I got this aspect `@Timed` in module A (with a target of `TimedAspect.class`), and I try to use `@Timed` on a method in module B. The generated proxy class in module b is:

```
@Proxy
public final class ServiceClass$Proxy extends ServiceClass {

  private Method callDownStream0;
  private MethodInterceptor callDownStream0Timed;

  public ServiceClass$Proxy(AspectProvider<Timed> timed, ApiClient api) {
    super(api);
    try {
      callDownStream0 = ServiceClass.class.getDeclaredMethod("callDownStream");
      callDownStream0Timed = timed.interceptor(callDownStream0, callDownStream0.getAnnotation(Timed.class));
    } catch (Exception e) {
      throw new IllegalStateException(e);
    }
  }

  @Override
  public byte[] callDownStream() {
    var call = new Invocation.Call<>(() -> super.callDownStream())
      .with(this, callDownStream0);
    try {
      callDownStream0Timed.invoke(call);
      return call.finalResult();
    } catch (InvocationException e) {
      throw e;
    } catch (Throwable e) {
    }
  }
}
```

Currently, module b fails to compile with
```
No dependency provided for io.avaje.inject.aop.AspectProvider<com.jojo.javalin.api.aspect.Timed> on 
com.jojo.javalin.shade.service.ServiceClass$Proxy
```
Currently the only way to get an annotation from module A to work with module B, is to add an explicit `@InjectModules(provides=TimedAspect.class)` to module A. 

One would expect that auto provide aspects would eliminate the need for explicit @InjectModules provides.

SentryMan:
Now modules can automatically provide aspects for use in other modules without directly exposing the underlying aspect target.

- Modifies the ExternalProvider to read the `autoProvidesAspects` from loaded modules
- adds `autoRequiresAspects` to Module Class
- updates DBeanScopeBuilder to read the autoRequiresAspects and use it for module ordering.

Fixes #269 (nice)

Now that I consider it, do we even need the target field in the `@Aspect` annotation anymore?

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/AspectAnnotationReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/AspectAnnotationReader.java
@@ -31,7 +31,7 @@ final class AspectAnnotationReader {
       if (aspect != null) {
         Meta meta = readTarget(anElement);
         if (meta != null) {
-          aspects.add(new AspectPair(anElement, meta.target, meta.ordering));
+          aspects.add(new AspectPair(anElement, meta.ordering));
         }
       }
     }
--- a/inject-generator/src/main/java/io/avaje/inject/generator/AspectMethod.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/AspectMethod.java
@@ -107,10 +107,10 @@ final class AspectMethod {
   }
 
   void writeSetupFields(Append writer) {
-    writer.append("  private Method %s;", localName).eol();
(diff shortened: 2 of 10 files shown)


rbygrave:
Note: These commits have been included in https://github.com/avaje/avaje-inject/pull/270

Issue #272, opened 2023-02-02
AutoRequires doesn't work when using mvn compiler plugin.


SentryMan:
It seems when you use inject with the compiler plugin, the `ExternalProvider` Serviceloader can't detect other modules.
```
<plugin>
  <groupId>org.apache.maven.plugins</groupId>
  <artifactId>maven-compiler-plugin</artifactId>
  <configuration>
    <annotationProcessorPaths>
      <path>
        <groupId>io.avaje</groupId>
        <artifactId>avaje-inject-generator</artifactId>
        <version>${avaje.inject.version}</version>
      </path>
    </annotationProcessorPaths>
  </configuration>
</plugin>
```

rbygrave:
huh, cause we can test that now with the extra blackbox-aspect module we have and I thought I had that configuration as I wanted to test how that went with java modules (module-info) and it all seemed good (I was checking that we didn't break modules when our autoProvides includes classes that were not strictly in the module-info exports ... and that was all fine. [the generated module, aspect implementation can all be internal to the module / not publically accessible to any other modules].

rbygrave:
Side note, when using java modules (module-info) we **_need_** to specify annotation processors via maven-compiler-plugin ... we can't just have them as provided scope dependencies.

SentryMan:
blackbox-aspect module does not use the AP in the mvn compile plugin

rbygrave:
I temporarily modified both blackbox-aspect  and blackbox-test-inject to use  mvn compiler plugin to specify the AP ... but I _ALSO_ at the same time changed them to use module-info.

I did that temporary to test out the module-info exports.

We could and probably should change both blackbox-aspect  and blackbox-test-inject to use  mvn compiler plugin to specify the AP permanently. I believe what we are saying here is that doesn't work and will reproduce this issue, so I'll try that.

rbygrave:
Our fix for this with maven is to include a new `avaje-inject-maven-plugin` into build / plugins like:

```xml
<plugin>
  <groupId>io.avaje</groupId>
  <artifactId>avaje-inject-maven-plugin</artifactId>
  <version>1.0</version>
  <executions>
    <execution>
      <phase>process-sources</phase>
      <goals>
        <goal>provides</goal>
      </goals>
    </execution>
  </executions>
</plugin>
```

What this does is generate 2 files in `target`:
- `target/avaje-module-provides.txt`
- `target/avaje-plugin-provides.txt`

[...]

PR #273, opened 2023-02-02
Show failing with compiler plugin

SentryMan:
#272

### DIFF
--- a/blackbox-test-inject/pom.xml
+++ b/blackbox-test-inject/pom.xml
@@ -42,14 +42,6 @@
       <version>1.7.1</version>
     </dependency>
 
-    <!-- annotation processor -->
-    <dependency>
-      <groupId>io.avaje</groupId>
-      <artifactId>avaje-inject-generator</artifactId>
-      <version>${project.version}</version>
-      <scope>provided</scope>
-    </dependency>
-
     <!-- test dependencies -->
     <dependency>
       <groupId>io.avaje</groupId>
@@ -73,5 +65,22 @@
     </dependency>
 
   </dependencies>
-
+	<build>
+		<plugins>
+			<!-- Add Compiler args for loom and errorprone -->
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<annotationProcessorPaths>
+						<path>
+							<groupId>io.avaje</groupId>
+							<artifactId>avaje-inject-generator</artifactId>
+							<version>${project.version}</version>
+						</path>
+					</annotationProcessorPaths>
+				</configuration>
+			</plugin>
+		</plugins>
+	</build>
 </project>


rbygrave:
FYI: blackbox-aspect is missing META-INF/services/io.avaje.inject.spi.Module ...

SentryMan:
it's there in the blackbox-aspect jar

rbygrave:
> it's there in the blackbox-aspect jar created by jar plugin

Ah yeah, I forgot that the annotation processor automatically generates that file for us.

WRT this issue, a "thing that works that we don't want to do" is to add the `blackbox-aspect` into the `<annotationProcessorPaths>` `<path>` like:
```
          <annotationProcessorPaths>
            <path>
              <groupId>io.avaje</groupId>
              <artifactId>avaje-inject-generator</artifactId>
              <version>${project.version}</version>
            </path>
            <path>
              <groupId>io.avaje</groupId>
              <artifactId>blackbox-aspect</artifactId>
              <version>${project.version}</version>
            </path>
          </annotationProcessorPaths>
```
... which is a strong indication that it's a classpath / module-path issue.

This makes me ponder that the path going forward might be to replace the service loading with ... find the module classes, find the TypeElement (of the external modules) ... find their methods ... read the `@DependencyMeta` from the methods.  (This is what the AP already does in order to support partial compiles, ... we approximately extend that to read the  `@DependencyMeta` from all the external inject modules as well.

So time for lunch and will ponder that ...

SentryMan:
It seems the classpath just isn't there during the mvn compiler plugin. I tried using Class.forName to try and manually get the module but it didn't work.

rbygrave:
> It seems the classpath just isn't there during the mvn compiler plugin

Yes I agree.  

This is suggesting to me that ServiceLoader just isn't going to work for us here ... need to instead look at reading the `@DependencyMeta` via the TypeElements etc.  Also means the ServiceLoader for Plugin also isn't going to work.

SentryMan:
It doesn't look like TypeElements work when the classpath isn't there. We may have to do something wacky with the class loading. Or else we can say it's our loss and accept that this will only work when we do it as a provided dependency.

SentryMan:
jsonb plugin works though, because jsonb is added to the cp by the generator AP

SentryMan:
oh wait, this also means that jsonb imports won't work on dependencies if using the mvn compiler plugin.

SentryMan:
or alternatively, I'm actually panicking for no reason, because type elements do in fact work without the cp. Jsonb is fine. I'm still having trouble loading the generated module classes with type elements though

rbygrave:
I have a plan of creating a maven plug in that prior to the compile phase
reads the classpath and writes a file with the class names of the other
modules in the classpath. Then ExternalProvides can read that file.

SentryMan:
> I have a plan of creating a maven plug in that prior to the compile phase reads the classpath and writes a file with the class names of the other modules in the classpath. Then ExternalProvides can read that file.

ok

PR #274, opened 2023-02-05
Auto-Require Maven Plugin

SentryMan:
First time I've ever made a maven plugin.
 
- Use serviceloader in mvn plugin to create txt files containing the provided/autoprovided classes
- update processor to read the text files and add to plugins/modules respectively (I think we can keep the existing service-loader stuff in there for stuff like `Jsonb` or other annotation processors that use plugins/modules).

Solves #272  #273 

(Doesn't build in the pipeline because plugin isn't deployed, but it works on local)


SentryMan:
Yeah, I'm good with this. I suck at naming things though, so feel free to change the names of things

rbygrave:
Great work. I have renamed the plugin to `avaje-inject-maven-plugin`.

My thought was to generate into target/classes something like target/classes/META-INFO/avaje-inject-provides.txt ... but I like what you've done here just putting into into target as it truely isn't needed in the final artifact etc.

I've released the `io.avaje : avaje-inject-maven-plugin : 1.0` with a goal of `provides`
and released the rest as `8.12-RC4`

Cheers, Rob.

Issue #276, opened 2023-02-06
Ideally avaje-inject-generator should not depend on avaje-inject


agentgt:
A lot of folks are not aware of this but you should almost never use the actual annotation classes in an annotation processor even your own. BTW I only know of couple libraries that generally do this correctly: MapStruct, JStachio (mine) and I believe Micronaut (so don't feel bad). It is also probably why no one knows about Hickory (or the new MapStruct gemtools).

So why is it problematic? Well the annotation processor is not guaranteed to have classes loaded and if your annotation references other classes (e.g. has a `Class<?>` parameter ) that are not yours you will have issues especially and I mean especially if you are dealing with modularized projects.

So why have you guys not run into issues? Well because both Maven and I think gradle as well will simply move whatever dependencies that are in `<annotationProcessorPaths>` into the classpath during compilation. Otherwise all libraries that are referenced with `requires` (static or not) in `module-info` (and not in the annotationProcessorPaths) will be put in the module-path.

The above is one the reasons why `<annotationProcessorPaths>` is required for modular projects but it is seriously limited and technically not correct either: https://issues.apache.org/jira/browse/MCOMPILER-391 and https://issues.apache.org/jira/browse/MCOMPILER-412

Otherwise if avaje-inject-generator does not have dependencies just having avaje-inject-generator on the classpath will work and you would not need to do `<annotationProcessorPaths>` for modular projects and this is one of the big advantages to doing it the painful. BTW this is also why lots of annotation processors shade their deps but you can't really do that for annotations.

So how do you access your own annotations without using them directly? You have to load the TypeElement.

https://github.com/avaje/avaje-inject/blob/12eba7f9f38ca9179afa0c47c4b149dbf6f0d503/inject-generator/src/main/java/io/avaje/inject/generator/Processor.java#L118

Instead of `Scope.class` you are going to pass a TypeElement. To get the TypeElement you can either get it from the call process function or just call `Elements.getTypeElement`: https://docs.oracle.com/en/java/javase/11/docs/api/java.compiler/javax/lang/model/util/Elements.html#getTypeElement(java.lang.CharSequence)

Anyway that is easy and fine however it becomes a pain in the ass to get annotation parameters. That is where Hickory comes in (or mapstruct gemtools but you would have to shade it).
https://javadoc.io/static/com.jolira/hickory/1.0.0/net/java/dev/hickory/prism/package-summary.html

[...]

PR #277, opened 2023-02-07
Use Avaje Prisms instead of referencing annotations directly

SentryMan:
For auto requires, you'll need to use the inject maven plugin though.

- Use Avaje(Hickory) prisms to access annotation information 
- Only service load modules/plugins if inject is on the classpath
- replace that one Stack with a Deque

takes care of #276


rbygrave:
Just to say I wouldn't invest too much time into this as I'm likely to go the path of using TypeElement / TypeMirror / AnnotationMirror (as I mentioned earlier).

SentryMan:
I mean I got it working, and it works when I modularize my javalin project. But it is what it is I suppose.

SentryMan:
> Just to say I wouldn't invest too much time into this as I'm likely to go the path of using TypeElement / TypeMirror / AnnotationMirror (as I mentioned earlier).

Sure you won't change your mind? hickory seems pretty rad and easy to use compared to AnnotationMirrors.

rbygrave:
Maybe. I'll put up a branch and we can compare ... noting that in the back of my mind there is:

- This library is expected to live for a long time / many years longevity / think 15 years 
- I need to 'own' it _including dependencies_ like hickory and I can't assume people will help over the maintenance lifetime

SentryMan:
> think 15 years
hmm, I would have thought that number would be higher. 

> I can't assume people will help over the maintenance lifetime
I like to stay optimistic about these sorts of things. One could even argue that having hickory makes the code easier to parse when compared to Annotation Mirrors and so would make it more likely for people to contribute.

In any case, if it breaks something in the future, my pride won't allow me to leave it broken if I was the one who caused it. I'd be the first to remove it.

agentgt:
@rbygrave Hickory is highly stable but just not supported. It is a very simple project and I was planning on forking it and releasing an updated version to maven central.

It is important to understand that Hickory is never a true dependency. It is a zero dep code generator. The annotations it use are retention = SOURCE. The code it produces has zero references to anything hickory. In fact it inspired me to make my own annotation library produce zero dep code as an option: https://jstach.io/jstachio/#maven_zerodep

MapStruct has gemtools which is a more update to date fork but sadly unlike Hickory it is not zero dep: https://github.com/mapstruct/tools-gem/issues/11

If @SentryMan and I did not use it the code change would not have happened in a couple of hours like we did yesterday.

Of course avaje instead of me could host a fork of hickory and then you would guarantee its future for these libraries. I'm planning to do it for jstachio at some point this month.

SentryMan:
Looks like prisms are back on the menu boys

rbygrave:
This was merged in via a squashed commit - https://github.com/avaje/avaje-inject/commit/064d8be4c47daeb51c4bbc5c03e1aed1f07e5403

rbygrave:
Closing as merged via the squashed commit - https://github.com/avaje/avaje-inject/commit/064d8be4c47daeb51c4bbc5c03e1aed1f07e5403 , these changes will be in 8.12

PR #278, opened 2023-02-08
Modify avaje-inject-generator to not use avaje-inject annotations directly

rbygrave:
- Make avaje-inject and optional dependency of avaje-inject-generator
- Use AnnotationMirror to read annotation attributes
- Add DetectClassPath to detect is ServiceLoader can be used to find externally provided dependencies


SentryMan:
As expected, it works fine. The only lingering doubt I have is that I feel it would make future alterations to the annotations somewhat tedious when compared to just generating prisms and using the values.

Also, do the ExternalProviderLoad and ExternalPluginProviderLoad really need to be separate classes? They are pretty close in functionality.

rbygrave:
Closing this as we went with https://github.com/avaje/avaje-inject/pull/277 and avaje-prisms

Issue #290, opened 2023-02-23
Beans that are AutoCloseable by instance are not closed.


agentgt:
For various reasons we often have a factory that will make a bean whose primary interface we want to expose but the concrete class also implements `AutoCloseable`. It is best probably to show this by example:

`HikariDataSource` is a special `DataSource` in that it also implements `AutoCloseable`. This is natural because its pool needs to be shutdown. 

However our factory method `@Bean` returns `DataSource` and not the specific `HikariDataSource` so I assume the generated code does not know this DataSource needs to be closed.

We do this in several other services as well.

It seems like the lifecycle handling of `BeanScope` should `instanceof` all instantiated beans to see if they are  closeable and close them. That or perhaps some configuration.

I can't recall if Spring does this but I think other DIs do (our in house manual DI does).

SentryMan:
Currently, the closable detection works at compile time by reading the actual return type of the factory method/singleton. 

> However our factory method @Bean returns DataSource and not the specific HikariDataSource

Might I ask why? If you return the specific HikariDataSource it's still registered as a DataSource in the scope and will have the autoclose functionality. Your other beans can still wire using the primary DataSource interface. 

Even so, I'll see if I can make it work at runtime.

rbygrave:
So with the PR just merged we can do:

```java
@Factory
class MyFactory  {

  @Bean(autoCloseable=true)
  DataSource provideDataSource(...) {
    return ...
  }

```
And then the generated code will add a `builder.addAutoClosable($bean)` after the bean is registered like:
```java
      var $bean = builder.register(bean);
      builder.addAutoClosable($bean);
```
... and if the instance is AutoCloseable then that will register with PreDestroy.


**Background:**
The thinking here is that this issue of need to do `bean instanceOf AutoCloseable` is restricted to beans that are created via `@Factory` `@Bean`.  The `@Bean` already has the `initMethod` and `destroyMethod` attributes (which are also common with Spring DI and Micronaut DI) to register lifecycle methods.  

So the thinking here is that the new `autoCloseable` attribute on `@Bean` is reasonably in line with the different but related `destroyMethod` attribtue - they are the 2 ways to tell the DI that we have a PreDestroy method that we want invoked when the DI context is closed.

agentgt:
> The thinking here is that this issue of need to do bean instanceOf AutoCloseable is restricted to beans that are created via @Factory @Bean. 

I agree and am massively pleased with the changes.

[...]

PR #292, opened 2023-02-23
Add Closeable Option to Factory Beans

SentryMan:
I prefer doing things at compile time if possible, so I'm still not 100% sold on this. Related to #290

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
@@ -165,7 +165,7 @@ final class BeanReader {
    * Return true if lifecycle via annotated methods is required.
    */
   boolean hasLifecycleMethods() {
-    return (postConstructMethod != null || preDestroyMethod != null || typeReader.isClosable());
+    return (postConstructMethod != null || preDestroyMethod != null);
   }
 
   List<MetaData> createFactoryMethodMeta() {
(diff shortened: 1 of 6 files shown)


rbygrave:
As an alternative to this ... add an explicit attribute to `@Bean` like `@Bean(autoClosable=true)` and when this is set to true then generate the extra check 

```java
@Factory 
MyFactory {

  @Bean(autoCloseable=true) // if the instance is AutoClosable register it with PreDestory
  DataSource datasource() {
   ...
  }

}

```

So instead of the usual:
```java
    if (builder.isAddBeanFor(DataSource.class)) {
      var factory = builder.get(MyFactory.class);
      var bean = factory.data();
      builder.register(bean);
    }
```


Generate code to that sends the registered bean to a `addPreDestroy($bean)` method (that takes Object and does the instance of check)
```java
    if (builder.isAddBeanFor(DataSource.class)) {
      var factory = builder.get(MyFactory.class);
      var bean = factory.data();
      builder.addPreDestroy($bean);
    }
```

The extra generated code with `@Bean(autoClosable=true)` is the ... 
```java
    
      var $bean = 
      builder.addPreDestroy($bean);
    
```

SentryMan:
ok I can roll with that

rbygrave:
I extended this PR via https://github.com/avaje/avaje-inject/pull/296 ... to move the if block and `instance of AutoCloseable` check into a new `builder.addAutoCloseable()` method.

rbygrave:
Some the changes in this PR are effectively included in https://github.com/avaje/avaje-inject/pull/296 but I squashed the commit when I merged it into master.

We can close this with the changes here being merged into master.

PR #301, opened 2023-02-28
Support Wiring Regular Maps

### LINKED ISSUE #298: Map Beans not detected
suppose I have a factory that provides

```java
  @Bean
  Map<String, Model> map() {

    return Map.of("hi",  new Model, ....);
  }
```

When I rely on this bean somewhere else.

```java
  @Bean
  SomeOtherBean otherBean( Map<String, Model> s) {
....
  }
```
I get this exception.

[ERROR] Dependencies [Model] are not provided - missing @Singleton, @Component, @Factory/@Bean or specify external dependency via @InjectModule requires attribute

One would expect that the map from the factory would be detected.

SentryMan:
Well, I didn't know that map wiring was originally only for getting qualified beans. Seems like a pretty uncommon use case compared to wiring regular maps.

- adds @QualifiedMap annotation to make generated DI classes have the old behavior
- modifies util methods to use regular `builder.get` unless the Map has the annotation.,

fixes #298

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/FieldReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/FieldReader.java
@@ -14,14 +14,16 @@ final class FieldReader {
   private final GenericType type;
   private boolean requestParam;
   private String requestParamName;
+  private final boolean isBeanMap;
 
   FieldReader(Element element) {
     this.element = element;
     this.name = Util.getNamed(element);
     this.nullable = Util.isNullable(element);
     this.utype = Util.determineType(element.asType());
-    this.fieldType = Util.unwrapProvider(utype.rawType());
-    this.type = GenericType.parse(utype.rawType());
+    this.isBeanMap = QualifiedMapPrism.isPresent(element);
+    this.fieldType = Util.unwrapProvider(utype.rawType(isBeanMap));
+    this.type = GenericType.parse(utype.rawType(isBeanMap));
   }
 
   boolean isGenericParam() {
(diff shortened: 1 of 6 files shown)


rbygrave:
> adds @QualifiedMap annotation to make generated DI classes have the old behavior

Yes. We note that this makes this a behavior change rather than a bugfix.  Existing code expecting `@QualifiedMap` behavior is broken / needs to change to get back the prior behavior. This is all good if the existing code barfs with something indicating a missing map dependency - hopefully we see a good error and people not reading the release notes don't get confused.

With having the 2 types of Map ... we ask which is the expected "default behavior" / should it be `@QualifiedMap` or the other way around? For me, I think that this way around makes sense in that it is the "qualified map injection" that is "a bit special" and it's more that avaje-inject didn't support injecting "normal maps" before ... so for me I think this way around is the correct long term thing to do.

Just need to check if the error existing code will see is good (as people frequently don't check release notes) ... or consider major version bump on this one.

SentryMan:
The error that'll show will be something like this 

[ERROR] Dependencies [java.util.Map<java.lang.String,org.example.coffee.list.Somei>] are not provided - missing @Singleton, @Component, @Factory/@Bean or specify external dependency via @InjectModule requires attribute

rbygrave:
Yup cool. I think we release `8.13` ... and then merge this into version `9.0`  (or `10.0` ... currently 7.x and 9.x would be versions for `javax` but we could change to use a suffix for the javax versions so release versions like  `8.13-javax`, `9.0-javax` etc.

SentryMan:
> we could change to use a suffix for the javax versions so release versions like 8.13-javax, 9.0-javax etc.

I'm thinking for the sake of consistency, we can have a final release for 7.13, but for 9.x onwards add the javax suffix.

rbygrave:
> we can have a final release for 7.13, but for 9.x onwards add the javax suffix.

Yes, that is the plan. I'll look to release 8.13 | 7.13 shortly.

PR #302, opened 2023-02-28
Component Import Annotaion

SentryMan:
Adds an annotation to import types and generate DI classes for them.


rbygrave:
Yeah, this is a very cool idea - love it !!

PR #309, opened 2023-03-02
Return all qualified beans from BeanScope map method

### LINKED ISSUE #289: Support Qualified Lists/Set
Suppose we want a list of qualified beans.
```
@Qualifier
@Retention(RetentionPolicy.RUNTIME)
@Target(ElementType.TYPE_USE)
public @interface Listener {}


@Inject
public void registerListeners(Set<@Listener ListenerInterface> foundListeners) {
//whatever
}
```
Currently, this won't work.

SentryMan:
Before the map didn't actually contain all the qualified beans. (say you had an interface and wanted to get all the implementations annotated with a @Qualifier annotation, it wouldn't work)

Fixes #289 (because you can get the list of qualified beans from a qualified map)

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/UtilType.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/UtilType.java
@@ -58,7 +58,12 @@ final class UtilType {
         return Util.extractList(rawType);
       case MAP:
         if (beanMap) {
-          return Util.extractMap(rawType);
+          var listType = Util.extractMap(rawType);
+          if (!listType.startsWith("java.util.List<")) {
+            throw new IllegalStateException(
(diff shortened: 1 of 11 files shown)


rbygrave:
>  it wouldn't work

I what way would it not work? The combination of Qualifier and Type should be unique apart from the issue of priority - `@Primary`, `@Secondary` ... but we wire unique instances by  Qualifier + Type.

So `Map<String, Type>` should be correct as I see it ... and I don't think we should expect multiple instances per Qualifier and hence not expect any use cases that expect or need multiple instances per qualifier?

SentryMan:
this doesn't actually fulfill @Tim203's request to get objects of disparate types

PR #310, opened 2023-03-04
ProxyBeanScope Can Call `listByAnnotation` While Scope is Being Built

SentryMan:
- Modify proxy scope to support listByAnnnotaion while scope is being built
- change the ordering of beans that require `BeanScope` (Will try to order as late as possible so that using the proxy scope is more useful)

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
@@ -99,10 +99,7 @@ final class BeanReader {
     if (constructor != null) {
       for (MethodReader.MethodParam param : constructor.params()) {
         Dependency dependsOn = param.dependsOn();
-        // BeanScope is always injectable with no impact on injection ordering
-        if (!dependsOn.dependsOn().equals(Constants.BEANSCOPE)) {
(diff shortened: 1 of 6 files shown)


rbygrave:
What is a "ProxyScope" ?
What is the use case justifying adding this? There isn't any test that comes with this PR that might explain the use case that this is trying to solve.

rbygrave:
> Modify proxy scope to support listByAnnnotaion while scope is being built

This looks like it does not provide a predictable result. That is, the result will vary depending the ordering of the wiring (which is controlled by types and qualifiers).

rbygrave:
> Will try to order as late as possible so that using the proxy scope is more useful

I think `@PostConstruct` should generally be used when something depends on the BeanScope. All PostConstruct methods fire after _everything_ is wired. For me I'm expecting BeanScope to be used in PostConstruct methods and it's not expected at least by me that an injected BeanScope should be used before then?

SentryMan:
> add listbyAnnotation injection using annotations, so went in this direction to get the same effect.

hmm, though I suppose encouraging reflection while creating beans is sort of the antithesis of a compile-time DI library. I'll table this for now unless demand increases.

PR #312, opened 2023-03-06
Add RequiresBean and RequiresProperty Annotations

SentryMan:
- adds RequiresBean and RequiresProperty
- modify bean/method readers to get the prisms and write the conditional statements
- bean types are automatically added to auto requires (to ensure module order is correct)
- modify builder and proxy scope to add `contains` methods
- no longer register null/`Optional.empty()` beans 

I'm thinking this is a good idea. Will not try to wire the bean if the condition is not met. This used in conjunction with `@Secondary` could be handy. 

```java

@Factory
@RequiresProperty(value="wire.configs", notEqualTo="false")
public class ConfigFactory {

  @Bean
  @RequiresBean(qualifiers="proxy") // some external bean we may provide manually or from another module 
  ProxyConfig proxyConfig() {

    return new ProxyConfig();
  }

  @Bean
  @Secondary // will default to this if not matched
  RealConfig config() {

    return new RealConfig();
  }
}

```
Generated proxyConfig factory method
```java

  public static void build_proxyConfig(io.avaje.inject.spi.Builder builder) {
    if (!builder.contains("proxy"))
     return;

    if (builder.isAddBeanFor(ProxyConfig.class, Config.class)) {
      var factory = builder.getNullable(ConfigFactory.class); 
      if(factory == null) return;
      var bean = factory.proxyConfig();
      builder.register(bean);
    }
  }
```


rbygrave:
To me I'm almost certain this is a copy of the equivalent spring feature and their string name matches to a full type name ... and we need to use the string rather than `Foo.class` as `Foo.class` isn't necessarily on the classpath (so we use the string literal like "org.example.Foo" rather than `Foo.class`).  I suspect this is the same feature but then the "proxy" literal in the example doesn't look right so maybe I don't understand this yet.

rbygrave:
As some background, as I see it **_IF_** the "conditional" in practice translates to ... _"wire these **10 beans** based on this **1 condition**"_ then custom scopes are a good fit.  **IF** the "conditional" in practice translates to ... _"wire this **1 bean** based in this **1 condition**"_ then yes `@ConditionalOnBean` would be a much better fit than a custom scope.

For me, the custom scopes case is the common one I have seen in practice although the good alternative would have been to use a `@ConditionalOnBean` on the Factory itself rather than each factory method meaning - all the beans this factory creates require that condition.

SentryMan:
It was pretty late when I wrote this so I guess I was unclear.

> their string name matches a full type name ... and we need to use the string rather than `Foo.class` as `Foo.class` isn't necessarily on the classpath (so we use the string literal like "org.example.Foo" rather than `Foo.class`). I suspect this is the same feature but then the "proxy" literal in the example doesn't look right so maybe I don't understand this yet.

well looks like I goofed, I somehow totally misinterpreted how the beanScope `contains(String type)` worked, and assumed that it worked based on the qualifier name (that is, `@Named` or `@Qualifier` annotations).

>  the good alternative would have been to use a `@ConditionalOnBean` on the Factory itself rather than each factory method meaning - all the beans this factory creates require that condition.

Oh, this PR does that too, it's why I changed the factory methods to use getNullable.

```
      var factory = builder.getNullable(ConfigFactory.class); 
      if(factory == null) return; // return if factory is not in scope (only possible when Factory has a conditionOnBean that doesn't match)
```

rbygrave:
> misinterpreted

I think this "name" area is confusing because Spring DI really does give beans a "name" (not just a qualifier) and that "name" is based on the class short name (TLDR Today this looks like a terrible idea). This was maybe a good idea way back in the day with xml wiring configuration and only a few beans being wired but imo its a terrible idea when we might be wiring hundreds of beans in a large multi-module application because using the class short name is prone to not being unique - its easy for a bean with a completely different type to override another.

Both Micronaut and avaje-inject both do NOT follow Spring DI here with respect to "bean name" and both only using Qualifier + Type. 

Depending on who we talk to "name" can be a confusing term and as you say we note `@Named` is really "_Qualifier_ name".

For myself, I'm thinking I should try and just talk about "Qualifier" and Type (and not use "name" anywhere - the `@ConditionalOnBean` annotation should use `type()` and not `name()`).

Note that Micronaut uses `@Requires` rather than the multitude of `@ConditionOn___` that Spring has. I preferred the `@Requires` when I last looked at this but I'll have another look and review.

SentryMan:
ok I can rename

SentryMan:
ah hang on, I'm inspired now

SentryMan:
I think I'll split requires into two, one for beans and another for properties.

SentryMan:
I'm also going to use `System.getProperty` for the property check, but will use avaje config if available

SentryMan:
Okay, @rbygrave how's this?

SentryMan:
```

@Factory
@RequiresProperty("useHttp")
public class HttpClientFactory {

  @Bean
  @RequiresBean(qualifiers = "someQualifier")
  Builder builder(
      Jsonb jsonb) {
...
  }

  @Bean
  @RequiresBean(missingBeans = ApiClient.class)
  ApiClient client(Builder b) {
...
  }
}

```


Generated class 

```
public final class HttpClientFactory$DI  {

  /**
   * Create and register HttpClientFactory.
   */
  public static void build(io.avaje.inject.spi.Builder builder) {
    if (Optional.getNullable(System.getProperty("useHttp")).isEmpty())
     return;

    if (builder.isAddBeanFor(HttpClientFactory.class)) {
      var bean = new HttpClientFactory();
      builder.register(bean);
    }
  }

  /**
   * Create and register Builder via factory bean method HttpClientFactory#builder().
   */
  public static void build_builder(io.avaje.inject.spi.Builder builder) {
    if (!builder.containsQualifier("someQualifier"))
     return;

    if (builder.isAddBeanFor(Builder.class)) {
      var factory = builder.getNullable(HttpClientFactory.class); 
      if(factory == null) return;
      var bean = factory.builder(builder.get(Jsonb.class,"!jsonb"));
      builder.register(bean);
    }
  }

  /**
   * Create and register ApiClient via factory bean method HttpClientFactory#client().
   */
  public static void build_client(io.avaje.inject.spi.Builder builder) {
    if (builder.contains(ApiClient.class))
     return;

    if (builder.isAddBeanFor(ApiClient.class)) {
      var factory = builder.getNullable(HttpClientFactory.class); 
      if(factory == null) return;
      var bean = factory.client(builder.get(Builder.class,"!b"));
      builder.register(bean);
    }
  }

}
```

SentryMan:
Hmm, using `System.getProperty` or avaje config feels like it would be inconsistent in practice. Do you think I should commit to using avaje config and throw illegalStateException when avaje config isn't found while trying to use`@RequiresProperty`?

SentryMan:
> As some background, as I see it **_IF_** the "conditional" in practice translates to ... _"wire these **10 beans** based on this **1 condition**"_ then custom scopes are a good fit. **IF** the "conditional" in practice translates to ... _"wire this **1 bean** based in this **1 condition**"_ then yes `@ConditionalOnBean` would be a much better fit than a custom scope.
> 

I haven't used custom scopes all that much, but say you're building an auto-configuration library using avaje, with a bunch of beans only supposed to be wired based on a bunch of separate conditions for multiple beans, I'm thinking using custom scopes would become little unwieldy.

SentryMan:
In any case, I think I got it working how I want. Please review when you get time.

inject-generator/src/main/java/io/avaje/inject/generator/Util.java
```diff
@@ -247,4 +250,134 @@ public static String trimMethod(String method) {
...
+  }
+
+  public static void buildPropertyConditional(
+      Append writer,
+      Set<String> containsProps,
+      Set<String> missingProps,
+      Map<String, String> propertyEquals,
+      Map<String, String> propertyNotEquals) {
+
+    final var propertyGetter = "Config.getOptional(\"%s\")";
```
rbygrave, line 312: I'm pretty sure we want to add 1 more level of abstraction.  A new interface (RequiresPropertyPlugin ?) that has methods:

- contains(String)
- equalTo(String, String)
- notEqualTo(String, String)


And then implement that RequiresPropertyPlugin using avaje Config and fallback to System properties.

inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
```diff
@@ -34,19 +34,20 @@ private ProcessingContext() {}
...
  static void init(ProcessingEnvironment env, Set<String> moduleFileProvided) {
     processingEnv = env;
     messager = processingEnv.getMessager();
     filer = processingEnv.getFiler();
     elementUtils = processingEnv.getElementUtils();
     typeUtils = processingEnv.getTypeUtils();
-
     ExternalProvider.registerModuleProvidedTypes(providedTypes);
     providedTypes.addAll(moduleFileProvided);
+    useAvajeConfig =  element("io.avaje.config.Config") != null;
```
rbygrave, line 50: I think we likely add an extra RequiresPropertyPlugin interface to give us 1 more level of indirection for the properties checks

inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
```diff
@@ -145,7 +202,7 @@ private String fullBuildMethod() {
   }
 
   String builderGetFactory() {
-    return String.format("      var factory = builder.get(%s.class);", factoryShortName);
+    return String.format("      var factory = builder.getNullable(%s.class); %n      if(factory == null) return;", factoryShortName);
```
rbygrave, line 205: I'm thinking this does not have to be nullable as we don't get here if any of the checks fail. I'll check this.

inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
```diff
@@ -243,12 +300,27 @@ void addImports(ImportTypeMap importTypes) {
...
     }
+    conditionTypes.forEach(importTypes::add);
+    missingTypes.forEach(importTypes::add);
+    if (!containsProps.isEmpty()
+        || !missingProps.isEmpty()
+        || !propertyEquals.isEmpty()
+        || !propertyNotEquals.isEmpty()) {
+      importTypes.add(
+          ProcessingContext.useAvajeConfig() ? "io.avaje.config.Config" : "java.util.Optional");
+    }
```
rbygrave, line 311: Refer the RequiresPropertyPlugin idea below

inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
```diff
@@ -96,6 +113,40 @@ public String toString() {
...
+    prism.value().forEach(t -> conditionTypes.add(t.toString()));
+    prism.missingBeans().forEach(t -> missingTypes.add(t.toString()));
+    qualifierNames.addAll(prism.qualifiers());
+  }
+
+  void processPropertyPrism(RequiresPropertyPrism prism) {
+
+    if (!ProcessingContext.useAvajeConfig()) {
+      throw new IllegalStateException(
+          "RequiresProperty functionality only works with Avaje Config");
```
rbygrave, line 126: Refer the RequiresPropertyPlugin idea below

inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
```diff
@@ -91,6 +141,17 @@ BeanReader read() {
...
     }
+
+    conditionTypes.forEach(importTypes::add);
+    missingTypes.forEach(importTypes::add);
+    if (!containsProps.isEmpty()
+        || !missingProps.isEmpty()
+        || !propertyEquals.isEmpty()
+        || !propertyNotEquals.isEmpty()) {
+      importTypes.add(
+          ProcessingContext.useAvajeConfig() ? "io.avaje.config.Config" : "java.util.Optional");
```
rbygrave, line 152: Refer the RequiresPropertyPlugin idea below

inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
```diff
@@ -63,6 +80,39 @@ public String toString() {
...
+    prism.value().forEach(t -> conditionTypes.add(t.toString()));
+    prism.missingBeans().forEach(t -> missingTypes.add(t.toString()));
+    qualifierNames.addAll(prism.qualifiers());
+  }
+
+  void processPropertyPrism(RequiresPropertyPrism prism) {
+
+    if (!ProcessingContext.useAvajeConfig()) {
+      throw new IllegalStateException(
+          "RequiresProperty functionality only works with Avaje Config");
```
rbygrave, line 93: Refer the RequiresPropertyPlugin idea below

blackbox-test-inject/src/main/java/org/example/myapp/conditional/BirdFactory.java
```diff
@@ -0,0 +1,27 @@
...
+import io.avaje.inject.RequiresProperty;
+import io.avaje.inject.Secondary;
+
+@Factory
+@RequiresProperty(value = "factory", equalTo = "bird")
+public class BirdFactory {
+
+  @Bean
+  @NoKiwi
+  public BlueJay jay() {
```
rbygrave, line 17: Rob, check that the RequiresProperty is 'inherited' for all the methods of the factory.

rbygrave (commented):
I'll have a look at these points but looking really good. 

Maybe the main thing is that I think we should add a new interface like RequiredPropertiesPlugin that has our 3 property checking methods and can implement that via Config or via System properties (or someone can put in their own implementation).

inject/src/main/java/io/avaje/inject/spi/DBeanMap.java
```diff
@@ -59,13 +64,20 @@ private void addSuppliedBean(SuppliedBean supplied) {
   }
 
   void register(Object bean) {
-    DContextEntryBean entryBean = DContextEntryBean.of(bean, nextBean.name, nextBean.priority);
+    if (bean == null || Optional.empty().equals(bean)) {
```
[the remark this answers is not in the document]
rbygrave: At the moment the factory level conditions are not inherited by the methods ... hence the null factory at the moment

rbygrave:
The follow up PR is - https://github.com/avaje/avaje-inject/pull/315

SentryMan:
Nice, don't have access to modify that one so I'll pull it over here and add on to it

inject/src/main/java/io/avaje/inject/RequiresBean.java
```diff
@@ -0,0 +1,56 @@
...
+@Repeatable(RequireBeans.class)
+@Target({TYPE, METHOD, ANNOTATION_TYPE})
+public @interface RequiresBean {
+
+  /**
+   * Expresses that beans of the given types should be available in the {@link BeanScope}.
+   *
+   * @return the class types of beans to check
+   */
+  Class<?>[] value() default {};
```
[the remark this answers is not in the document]
SentryMan: perhaps, but the reason I took that out is that I thought there would be some inconsistency with how the modules get ordered, since unlike the Class example we can't define a soft dependency and auto require on a class not in the cp

inject/src/main/java/io/avaje/inject/RequireConfigs.java
```diff
@@ -0,0 +1,15 @@
...
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.TYPE, ElementType.METHOD})
+@interface RequireConfigs {
```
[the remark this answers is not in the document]
SentryMan: oh, didn't know you could do that, you learn something new every day

inject-generator/src/main/java/io/avaje/inject/generator/Util.java
SentryMan: Yeah I like this, will add to avaje config

rbygrave:
Groovy. I think we merge this in and tweak in master (add javadoc etc)

inject/src/main/java/io/avaje/inject/spi/DBeanMap.java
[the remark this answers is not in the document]
rbygrave: Yeah, I'm now thinking we can add this check back in. As you say, is it possible.

PR #313, opened 2023-03-06
Update Javax/Jakarta Scripts to Add/Remove `-javax` in Parent Version.

SentryMan:
I for one welcome our new AI overlords.

- Will add/remove `-javax` from the parent version (will even work on RC versions)
- Now correctly replaces javax when going back to jakarta.
- update GitHub actions to test javax build.


rbygrave:
Very nice !!

PR #317, opened 2023-03-10
Now Scripts will Add/Remove `-javax` even if there is no `-SNAPSHOT`

SentryMan:
Brought to you by ChatGPT.

>In summary, this sed command uses regular expressions to match and replace the first occurrence of `<version>...</version>` in the file. If the original version contains the -SNAPSHOT suffix, it adds the -javax suffix before the -SNAPSHOT suffix. If the original version doesn't contain the -SNAPSHOT suffix, it adds the -javax suffix after the version number.

### DIFF
--- a/jakarta-to-javax.sh
+++ b/jakarta-to-javax.sh
@@ -2,13 +2,13 @@
 
 ## adjust pom dependencies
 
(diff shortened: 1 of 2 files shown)


SentryMan:
hmm wait a sec

SentryMan:
@rbygrave do you want the script to put -javax before or after `-RCX` ?

SentryMan:
if before, then you can merge this

PR #322, opened 2023-03-13
Make the avaje-inject dependency of avaje-inject-generator optional

rbygrave:


SentryMan:
This'll cause Jsonb plugin to not automatically work on gradle.

rbygrave:
Ah stink !!  :)

SentryMan:
it is what it is

SentryMan:
though if we figure out the gradle plugin we could take it out

PR #324, opened 2023-03-16
Expose PropertyPlugin from BeanScopeBuilder

SentryMan:
Okay so stay with me here, I'm thinking that we could expose the propertyPlugin from the builder and use that to help configure avaje plugins based on properties alone.

For example:

```java
public final class AvajePlugin implements io.avaje.inject.spi.Plugin {

  @Override
    return new Class<?>[]{SomeClass.class};
  }

  @Override

	 if( builder.propertyPlugin().contains("prop")) {
		// register something
	 };

  }
}
```


rbygrave:
Yeah cool, the Jsonb example seems like a better example but yeah I like it !!

PR #335, opened 2023-04-17
Aspect.Import Annotation

SentryMan:
Adds an annotation that marks the given annotation as an aspect.


rob-bygrave:
yeah, nice !!

PR #339, opened 2023-04-27
Go back to regular ServiceLoader

SentryMan:
Reverts avaje/avaje-inject#338 if we decide on the cyclical solution.


rob-bygrave:
Nice, thanks!!

PR #344, opened 2023-05-05
Add Invocation.invokeUnchecked() ... for ease of use when aspect isn't dealing wth checked exceptions

SentryMan:
for my aspects it's kinda annoying to deal with checked exceptions. 
```
 SVUtil.createScopedMDC(
          mdc,
          () -> {
            try {
              return invoke.invoke();
            } catch (final Throwable e) {
              if (e instanceof final RuntimeException ex) throw ex;
            }
      });
```
Sure would be nice to have a method I could call when I know my aspect isn't dealing with checked exceptions.


rob-bygrave:
interface default methods for the win !!

PR #345, opened 2023-05-05
Fix Throwable Aspects

SentryMan:
Currently, proxying a method that `throws Throwable` causes a compilation error.

- if thrown types are unchecked, don't write a superfluous catch clause
- if method throws Throwable, rethrow it.


rob-bygrave:
Nice, I forgot about this case.

PR #348, opened 2023-05-20
Check if Generic Type Exists Before Writing

### LINKED ISSUE #347: Generic type not generating correctly on second implementation
Hello, I have a little issue with the code generation for a generic type. this probably isn't a great description so here is an example

This is my project structure (grade): 
```
datasource:
 \_ common
    - public interface Repository<T, M> 
 \_ mongodb 
   - public abstract class MongoRepository<T> implements Repository<T, Document>
...

player-service
\_ server
    - public class PlayerMetaRepository extends MongoRepository<MetaEntry> { 
```

On compilation of player-service there is a syntax error with the PlayerMetaRepository$DI.

```
public final class PlayerMetaRepository$DI  {

 // error is here at "T"
  public static final Type TYPE_RepositoryTDocument = new GenericType<package.datasources.Repository<T,org.bson.Document>>(){}.type();

  /**
   * Create and register PlayerMetaRepository.
   */
    if (builder.isAddBeanFor("playermeta", PlayerMetaRepository.class, MongoRepository.class, TYPE_RepositoryTDocument)) {
      var bean = new PlayerMetaRepository(builder.get(MongoDBDatasource.class,"!datasource"));
      builder.register(bean);
    }
  }
}
```

Error:
```
error: cannot find symbol  
public static final Type TYPE_RepositoryTDocument = new GenericType<package.datasources.Repository<T,org.bson.Document>>(){}.type();
                                                                                                        ^
  symbol:   class T
  location: class PlayerMetaRepository$DI
```
It is referencing `Repository` instead of its implementation `MongoRepository`, and then the T is not being accounted for D:

Thank you for your time 🍎

SentryMan:
It seems when using class -> generic class -> generic class/interface, the type information is lost when calling `getInterfaces`/`getSuperClass`, with no way to retrieve it. 

To solve this, I have made it so that if an unknown type is added to the list of super interfaces we will write the raw type without generic parameters.

This isn't a problem when using class -> generic interface -> generic interface.

fixes #347

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
@@ -1,9 +1,14 @@
 package io.avaje.inject.generator;
 
-import static io.avaje.inject.generator.ProcessingContext.*;
+import static io.avaje.inject.generator.ProcessingContext.asElement;
+import static io.avaje.inject.generator.ProcessingContext.element;
+import static io.avaje.inject.generator.ProcessingContext.logWarn;
+import static io.avaje.inject.generator.ProcessingContext.types;
+
 import java.lang.annotation.Annotation;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.stream.Stream;
 
 import javax.lang.model.element.Element;
 import javax.lang.model.element.ElementKind;
@@ -127,11 +132,11 @@ final class TypeExtendsReader {
       extendsInjection.read(baseType);
     }
     readInterfaces(baseType);
-    TypeElement superElement = superOf(baseType);
+    final TypeElement superElement = superOf(baseType);
     if (superElement != null) {
       if (qualifierName == null) {
-        String baseName = baseType.getSimpleName().toString();
-        String superName = superElement.getSimpleName().toString();
+        final String baseName = baseType.getSimpleName().toString();
+        final String superName = superElement.getSimpleName().toString();
         if (baseName.endsWith(superName)) {
           qualifierName = baseName.substring(0, baseName.length() - superName.length()).toLowerCase();
         }
@@ -148,7 +153,7 @@ final class TypeExtendsReader {
   }
 
   private String initProvidesAspect() {
-    for (String providesType : providesTypes) {
+    for (final String providesType : providesTypes) {
       if (Util.isAspectProvider(providesType)) {
         return Util.extractAspectType(providesType);
       }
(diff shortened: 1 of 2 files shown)


inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
```diff
@@ -148,59 +155,73 @@ void process(boolean forBean) {
...
     }
     return "";
   }
 
-  private void addSuperType(TypeElement element) {
+  private void addSuperType(TypeElement element, TypeMirror mirror) {
     readInterfaces(element);
-    final String fullName = element.getQualifiedName().toString();
-    if (!fullName.equals(JAVA_LANG_OBJECT) && !fullName.equals(JAVA_LANG_RECORD)) {
+    final String fullName = mirror.toString();
```
rbygrave, line 168: What is the thinking behind this change to use the mirror.toString() ?

inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
```diff
@@ -148,59 +155,73 @@ void process(boolean forBean) {
...
+
+        extendsTypes.add(knownType ? type : genericType.topType());
+
         extendsInjection.read(element);
       }
-      addSuperType(superOf(element));
-    }
-  }
 
-  private TypeElement superOf(TypeElement element) {
```
rbygrave, line 172: It looks like the superOf() method got removed?

inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
```diff
@@ -148,59 +155,73 @@ void process(boolean forBean) {
...
     if (JAVA_LANG_OBJECT.equals(rawType)) {
     } else if (rawType.indexOf('.') == -1) {
       logWarn("skip when no package on interface " + rawType);
     } else if (Constants.AUTO_CLOSEABLE.equals(rawType) || Constants.IO_CLOSEABLE.equals(rawType)) {
       closeable = true;
     } else {
+      final var genericType = GenericType.parse(rawType);
       if (qualifierName == null) {
-        String mainType = GenericType.removeParameter(rawType);
+        final String mainType = genericType.topType();
```
rbygrave, line 208: What is the thinking behind this change here?

inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
rbygrave: Ah, using mirror .getSuperClas() .asElement()

inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
rbygrave: Ah so `element.getQualifiedName().toString()` loses the generic type info ...

SentryMan: exactly

inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
SentryMan: reusing the same generic type instance

rbygrave: No actual difference between the 2 here - no effective change

PR #351, opened 2023-05-25
Sanitize DI class imports

SentryMan:
On the discord, an issue was reported that something like this generates invalid adapters.

```
@Singleton
public class ServersService {
  ...
  public @Inject ServersService(final @NotNull JedisPool jedisPool, final @NotNull NatsConnection natsConnection) {
    this.jedisPool = jedisPool;
    this.natsConnection = natsConnection;
  }
  ...
```

the generated DI class has annotations in the imports causing compilation to fail.

I could not replicate the issue on my local, but I'm sure this'll fix it. (or else do absolutely nothing).

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/Util.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/Util.java
@@ -64,7 +64,7 @@ final class Util {
       return type.replace("[]", "");
     }
 
-    var start = pos == 0 ? type.substring(0, pos + 1) : "";
+    var start = pos == 0 ? type.substring(0, pos) : "";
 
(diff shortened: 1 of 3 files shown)


SentryMan:
It seems like it may be an IntelliJ thing. I can't really replicate since I don't have.

inject-generator/src/main/java/io/avaje/inject/generator/Util.java
```diff
@@ -58,6 +58,17 @@ public static String trimAnnotations(String type) {
     return type.substring(0, pos + 1) + type.substring(type.lastIndexOf(' ') + 1);
   }
 
+  public static String sanitizeimports(String type) {
```
rbygrave, line 61: missing tests for this

PR #354, opened 2023-06-01
Constant Empty Optional

SentryMan:
Instead of allocating a new optional empty for every bean.


rob-bygrave:
Nice!!  Thanks for removing those empty files ... not sure how I managed to leave them there :)

PR #355, opened 2023-06-17
Exclude Blackbox Module from Lift

SentryMan:
configure lift to skip blackbox module


rbygrave:
Why do we want this change?

SentryMan:
Because it's throwing off the bug reports? I mean I don't feel like fixing the sca issue of the black box test

rob-bygrave:
I can't find the Lift report detail that suggests this change (hence I don't know why it is suggesting this change).  Can you include that here?

This change does not make sense to me. It looks like a bad suggestion to me.

SentryMan:
Optional.empty doesn't create a new instance every time, so it's kinda redundant to define a second constant. Even so, I made this pr primarily to exclude the black box test module.

PR #359, opened 2023-07-02
Refactor Processor, extract methods and use stream.peek()

rbygrave:


inject-generator/src/main/java/io/avaje/inject/generator/Processor.java
```diff
@@ -143,6 +123,21 @@ public boolean process(Set<? extends TypeElement> annotations, RoundEnvironment
     return false;
   }
 
+  private static Set<TypeElement> importedElements(RoundEnvironment roundEnv) {
+    return roundEnv.getElementsAnnotatedWith(element(ImportPrism.PRISM_TYPE)).stream()
+        .map(ImportPrism::getInstanceOn)
+        .flatMap(p -> p.value().stream())
+        .map(ProcessingContext::asElement)
+        .peek(e -> addImportedType(e.getQualifiedName().toString()))
```
SentryMan, line 131: Last I checked `Stream.peek` has a tendency to skip over elements in certain situations

rbygrave: Hmm, I've used peek() a decent amount and not hit a skipping issue? You want it changed back?

SentryMan: better safe than sorry

PR #360, opened 2023-07-02
Refactor away from using  `Stream.peek()`

SentryMan:
https://rules.sonarsource.com/java/RSPEC-3864/


rbygrave:
> A key difference with other intermediate Stream operations is that the Stream implementation is free to skip calls to peek() for optimization purpose. 

Well yes but what the Javadoc says is:

> Javadoc: In cases where the stream implementation is able to optimize away the production of some or all the elements (such as with _short-circuiting operations like findFirst_, _or in the example described in count_), the action will not be invoked for those elements.


So to me ...

> This can lead to peek() being unexpectedly called only for some or none of the elements in the Stream.

This sounds more unexpected and scary that imo I believe it is.

That is, I'm pretty comfortable that `peek()` is never getting optimised away here as there is no short-circuit operation in play nor a special case like `count()`.  So for me, I'm happy using `peek()` here.

PR #361, opened 2023-07-03
Default implementations for redundant PropertyRequiresPlugin methods

Mechite:
Creates some default implementations returning the opposite of contains and equalTo for their opposing redundant methods.
I don't know why these are here, and perhaps they don't serve a purpose and their usages can be migrated to, well, the opposite of contains and equalTo.

### DIFF
--- a/inject/src/main/java/io/avaje/inject/spi/PropertyRequiresPlugin.java
+++ b/inject/src/main/java/io/avaje/inject/spi/PropertyRequiresPlugin.java
@@ -19,7 +19,9 @@ public interface PropertyRequiresPlugin {
   /**
    * Return true if the property is not defined.
    */
-  boolean missing(String property);
+  default boolean missing(String property) {
+    return !contains(property);
+  }
 
   /**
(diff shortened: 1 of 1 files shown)


rbygrave:
>  their usages can be migrated to, well, the opposite of contains and equalTo.

Technically this is valid - we could do that.

As I see it, the main reason why we would not do that and stick to `missing()` and `notEqualTo()` is because that can be deemed easier to read than the more subtle `!`.

Example: `missing()`
```java
    if (builder.contains(Kiwi.class)
      || builder.property().missing("finch-time")
      || builder.property().contains("secondary")) {
      return;
    }
```

vs `!`

```java
    if (builder.contains(Kiwi.class)
      || !builder.property().contains("finch-time")
      || builder.property().contains("secondary")) {
      return;
    }
```

So I think we merge this in - this is definitely an improvement.

There is a followup of if we stop using missing() and notEqualTo() and use `!` - is that too subtle / harder to read the generated code etc.

PR #368, opened 2023-07-12
BeanScope PostContruct

SentryMan:
Now can access the completed BeanScope from post construct methods
 ```
@PostConstruct
  void postCon(BeanScope scope) {

  }
  ```


rob-bygrave:
Nice !!

PR #370, opened 2023-07-16
Fix Module Debug Logging

SentryMan:
It seems that even when I add that swanky new adapter from slf4j:

```
		<dependency>
			<groupId>org.slf4j</groupId>
			<artifactId>slf4j-jdk-platform-logging</artifactId>
			<scope>runtime</scope>
		</dependency>
```

the debug log looks like this instead of showing the loaded modules: `22:20:44.627 [main] DEBUG io.avaje.inject - building with modules {0}`

Also adds logging for the detected modules from the generator.

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/ExternalProvider.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/ExternalProvider.java
@@ -1,6 +1,7 @@
 package io.avaje.inject.generator;
 
-import static io.avaje.inject.generator.ProcessingContext.*;
+import static io.avaje.inject.generator.ProcessingContext.logDebug;
+import static io.avaje.inject.generator.ProcessingContext.logWarn;
 
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
@@ -36,24 +37,23 @@ final class ExternalProvider {
       return;
     }
 
-    for (final Module module :
-        ServiceLoader.load(Module.class, ExternalProvider.class.getClassLoader())) {
-      try {
-        logDebug("Loaded External Module %s", module.getClass().getCanonicalName());
+    for (final Module module:ServiceLoader.load(Module.class, ExternalProvider.class.getClassLoader())){
+    try {
+      logDebug("Loaded External Module %s", module.getClass().getCanonicalName());
 
-        for (final Class<?> provide : module.provides()) {
-          providedTypes.add(provide.getCanonicalName());
-        }
-        for (final Class<?> provide : module.autoProvides()) {
-          providedTypes.add(provide.getCanonicalName());
-        }
-        for (final Class<?> provide : module.autoProvidesAspects()) {
-          providedTypes.add(Util.wrapAspect(provide.getCanonicalName()));
-        }
-      } catch (final ServiceConfigurationError expected) {
-        // ignore expected error reading the module that we are also writing
+      for (final Class<?> provide : module.provides()) {
+        providedTypes.add(provide.getCanonicalName());
       }
+      for (final Class<?> provide : module.autoProvides()) {
+        providedTypes.add(provide.getCanonicalName());
+      }
+      for (final Class<?> provide : module.autoProvidesAspects()) {
+        providedTypes.add(Util.wrapAspect(provide.getCanonicalName()));
+      }
+    } catch (final ServiceConfigurationError expected) {
+      // ignore expected error reading the module that we are also writing
     }
+  }
   }
 
   /**
@@ -65,6 +65,7 @@ final class ExternalProvider {
       return;
     }
     for (final Plugin plugin : ServiceLoader.load(Plugin.class, Processor.class.getClassLoader())) {
+      logDebug("Loaded Plugin %s", plugin.getClass().getCanonicalName());
       for (final Class<?> provide : plugin.provides()) {
         defaultScope.pluginProvided(provide.getCanonicalName());
       }
--- a/inject-generator/src/main/java/io/avaje/inject/generator/ExternalProvider.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/ExternalProvider.java
@@ -3,6 +3,7 @@ package io.avaje.inject.generator;
 import static io.avaje.inject.generator.ProcessingContext.logDebug;
 import static io.avaje.inject.generator.ProcessingContext.logWarn;
 
+import java.util.Iterator;
 import java.util.ServiceConfigurationError;
 import java.util.ServiceLoader;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


Mechite:
I think it's better if we keep using MessageFormat though since it usually reads a little bit better:

Instead of this diff:
```diff
- log.log(INFO, "Wired beans in {0}ms", (System.currentTimeMillis() - start));
+ log.log(INFO, String.format("Wired beans in %sms", System.currentTimeMillis() - start));
```

we can have this:
```diff
- log.log(INFO, "Wired beans in {0}ms", (System.currentTimeMillis() - start));
+ log.log(INFO, MessageFormat.format("Wired beans in {0}ms", System.currentTimeMillis() - start));
```

Though, we could also commit to the swanky new adapter and make it use `MessageFormat` so everyone gets that fixed up. In theory we can say it's a bug.

Actually, that "swanky new adapter" seems to be really similar to https://github.com/avaje/slf4j-jdk-platform-logging/, except there is a `SLF4JPlarformLoggerFactory` (note the typo, Plarform). Not sure where the slf4j repo for it is, but I saw that extra file in the sources JAR.

rob-bygrave:
Yeah, this was logged as https://jira.qos.ch/browse/SLF4J-545 which is a duplicate of https://jira.qos.ch/browse/SLF4J-529 and has the related PR of https://github.com/qos-ch/slf4j/pull/273 which is as yet not merged - yikes.

rob-bygrave:
So to be clear, there is a bug in SLF4J here and the logging message changes here are not ideal.

There is an alternative "fix" provided by the fact that avaje libs do NOT use System.Logger directly but instead obtain the System.Logger via `AppLog` like:.

```java
private static final System.Logger log = AppLog.getLogger("io.avaje.inject");
```
See https://github.com/avaje/avaje-applog

Now the original reason for using AppLog.getLogger() and not using System.getLogger() directly was due to AWS Lambda runtime not actually honouring the service loading of a System.Logger implementation. So I added this 1 extra level of indirection to allow controlling the System.Logger implementation for the AWS Lambda runtime but also it felt like a good idea to have this.

This means we are alternatively use `avaje-applog-alf4j` - https://github.com/avaje/avaje-applog-slf4j/tree/main and then all the AppLog SystemLoggers use the logger it provides and yes that does the correct MessageFormat.

**TLDR**: The alternative (until the SL4J bug fix is merged) is to add `avaje-applog-alf4j` and then any AppLog System.Loggers will log correctly.

SentryMan:
> TLDR: The alternative (until the SL4J bug fix is merged) is to add avaje-applog-slf4j and then any AppLog System.Loggers will log correctly.

Personally, I find it inconvenient that I have to add a whole dependency just for debugging purposes. Are you sure we can't just format it ourselves using MessageFormat?

Mechite:
I guess until they fix the issue it is best we don't merge this.
For the time being, as I have some software to deploy and it would be preferred to have things like this fixed, I've settled on this solution (anyone using JitPack can copy this into their POM for a temporary fix).
```xml
<!--
TODO - Revert to this dependency once the MessageFormat changes are accepted.
<dependency>
    <groupId>org.slf4j</groupId>
    <artifactId>slf4j-jdk-platform-logging</artifactId>
    <version>${slf4j.version}</version>
    <scope>runtime</scope>
</dependency>
-->
<dependency>
    <groupId>com.github.Mechite.slf4j</groupId>
    <artifactId>slf4j-jdk-platform-logging</artifactId>
    <version>patch-1-SNAPSHOT</version>
    <scope>runtime</scope>
</dependency>
```

SentryMan:
> I guess it is best we don't merge this as they will eventually fix the issue. 

The PR was raised in 2021, so I'm doubting hard.

Mechite:
> > I guess it is best we don't merge this as they will eventually fix the issue.
> 
> The PR was raised in 2021, so I'm doubting hard.

I guess given this then, while it's controversial, I think it's a better idea to just format things ourselves anyway since at the end of the day even if the whole codebase ended up littered with MessageFormat.format calls, this is the most resilient way to do it (the logging will continue to work on every single version of the slf4j adapter).

rob-bygrave:
> Are you sure we can't just format it ourselves using MessageFormat?

As I see it, its the "thin end of the wedge" in that **_ALL_** the avaje libraries and **_ALL_** the ebean libraries all use AppLog SystemLogger and they hence ALL expect the correct message format to "work".

For me, it isn't just these few log messages in avaje-inject that will not work as expected.


> com.github.Mechite.slf4j : slf4j-jdk-platform-logging

Nice.  If we want that in maven central via io.avaje we can do that.


> The PR was raised in 2021, so I'm doubting hard.

Yeah. Probably need to ask why it hasn't been merged yet.

SentryMan:
> For me, it isn't just these few log messages in avaje-inject that will not work as expected.

Figuring out this issue actually shed light on why config wasn't logging correctly either. Indeed, the message format doesn't work anywhere where slf4j bridges

Mechite:
> > Are you sure we can't just format it ourselves using MessageFormat?
> 
> As I see it, its the "thin end of the wedge" in that **_ALL_** the avaje libraries and **_ALL_** the ebean libraries all use AppLog SystemLogger and they hence ALL expect the correct message format to "work".
> 
> For me, it isn't just these few log messages in avaje-inject that will not work as expected.
> 
> > com.github.Mechite.slf4j : slf4j-jdk-platform-logging
> 
> Nice. If we want that in maven central via io.avaje we can do that.
> 
> > The PR was raised in 2021, so I'm doubting hard.
> 
> Yeah. Probably need to ask why it hasn't been merged yet.

```java
public class FormattedLogger implements System.Logger {

    private final System.Logger delegate;

    public FormattedLogger(System.Logger delegate) {
        this.delegate = delegate;
    }

    (...) // other methods excluded for brevity, they would forward to delegate

    @Override
    public void log(Level level, ResourceBundle bundle, String format, Object... params) {
        delegate.log(level, bundle, MessageFormat.format(String.format(format, params), params), params);
    }
}
```
What you can do is simply make a wrapping logger that formats things automagically.

This implementation does `String.format` as well as `MessageFormat.format` meaning by the time the log is passed through, it would have already been formatted in every possible way and the params are simply ignored by the underlying logger as there is nothing left to format away.

rob-bygrave:
> This implementation does String.format as well as MessageFormat.format meaning 

As I see it, it formats too early - before the check if it needs to log at all is performed. That might be ok for a lot of cases but not ideal for any "hot conditional debugging" type cases as I see it.

rob-bygrave:
> As I see it, it formats too early

_But_ ... that could be accounted for I think by just adding an IF block - hmm, interesting thought ...

SentryMan:
I guess this is more of an SLF4J problem

Mechite:
Refer to https://github.com/avaje/avaje-inject/issues/373.

PR #372, opened 2023-07-17
Consumer Posthook VarArgs

SentryMan:
The asymmetry is really starting to get to me

### DIFF
--- a/inject/src/main/java/io/avaje/inject/BeanScopeBuilder.java
+++ b/inject/src/main/java/io/avaje/inject/BeanScopeBuilder.java
@@ -221,7 +221,7 @@ public interface BeanScopeBuilder {
   BeanScopeBuilder addPostConstructHooks(Runnable... runnables);
 
   /**
-   * Adds hook that will execute after this scope is built.
+   * Adds hooks that will execute after this scope is built.
    *
    * @param consumers the PostConstruct hooks to run after the BeanScope is constructed
    */
--- a/inject/src/main/java/io/avaje/inject/BeanScopeBuilder.java
+++ b/inject/src/main/java/io/avaje/inject/BeanScopeBuilder.java
@@ -225,7 +225,8 @@ public interface BeanScopeBuilder {
    *
    * @param consumers the PostConstruct hooks to run after the BeanScope is constructed
    */
-  BeanScopeBuilder addPostConstructHooks(BeanScopeConsumer... consumers);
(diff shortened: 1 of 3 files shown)


inject/src/main/java/io/avaje/inject/BeanScopeConsumer.java
```diff
@@ -0,0 +1,6 @@
+package io.avaje.inject;
+
+import java.util.function.Consumer;
+
+@FunctionalInterface
```
rob-bygrave, line 5: Need some javadoc here?

SentryMan: I mean the name is pretty descriptive

Mechite:
Wouldn't this result in heap pollution (since its varargs with generics)?

SentryMan:
nah, because we create a separate non-generic interface that extends `Consumer`.

SentryMan:
> use @SafeVarargs

`@SafeVarargs annotation cannot be applied to non-final instance method addPostConstructHooks`

I'm pretty sure what I have already should work.

SentryMan:
Tried on my local, and I get compiler warnings everywhere I use the postHook method, I think I'll stick with the interface approach

inject/src/main/java/io/avaje/inject/BeanScopeBuilder.java
```diff
@@ -216,16 +216,16 @@ default <D> BeanScopeBuilder provideDefault(Type type, Supplier<D> provider) {
...
 
   /**
-   * Adds hook that will execute after this scope is built.
+   * Adds hooks that will execute after this scope is built.
    *
-   * @param consumer the PostConstruct hook to add to the bean scope
+   * @param consumers the PostConstruct hooks to run after the BeanScope is constructed
    */
-  BeanScopeBuilder addPostConstructConsumerHook(Consumer<BeanScope> consumer);
+  BeanScopeBuilder addPostConstructHooks(BeanScopeConsumer... consumers);
```
rbygrave, line 228: I suspect this is only ever called with 1 consumer.  I suspect the same for the Runnable hooks.

... so I _think_ these should both be changed to not use var-args ?

SentryMan: Oh yeah, we could also do that

Issue #304, opened 2023-02-28
Custom scope is ignored for Factory Beans


Tim203:
```java
@Factory
public class AFactory {
  @Bean
  @ACustomScope
  Hello hello() {}

  @Bean
  @BCustomScope
  ExtendsHello extendsHello() {}
}
```
(where ExtendsHello extends Hello)
Both are registered regardless whether any of the scopes are added as a module.

This leads to problems e.g. where you'd expect one binding of Hello, but instead there are two registered: Hello & ExtendsHello

rbygrave:
When we put the `@BCustomScope` on the AFactory itself we should get the result we are after - and this is what I'd expect assuming all the beans the factory creates are expected to be in the custom scope.

So:

```java
@BCustomScope
@Factory
public class AFactory {
  @Bean
  Hello hello() {}

  @Bean
  ExtendsHello extendsHello() {}
}


```

SentryMan:
He has 2 custom scopes. Depending on which custom module he adds he wants to wire a certain factory bean.

SentryMan:
![image](https://user-images.githubusercontent.com/32279667/221969647-260920e1-5d45-44ec-8dc9-7e10e57429b5.png)

rbygrave:
Mike is a big user of custom scopes in a "build application". For that application there is a base custom scope for "all platforms" and then custom scopes for windows, linux, mac etc ... the custom scopes form a hierarchy.  When the build application starts it determines the platform it's building for and then uses the appropriate custom scopes. For example, when building on linux the windows custom scope is not used at all.

The question here is, **_why isn't `@BCustomScope` on the factory_** rather than putting it on each bean.  Putting the scope on each `@Bean` implies the factory _could_ produce beans for different scopes and that isn't something we want - we want the factory to only build beans for a single scope.

SentryMan:
> we want the factory to only build beans for a single scope.

alrighty then

Tim203:
> The question here is, **_why isn't `@BCustomScope` on the factory_** rather than putting it on each bean. 

The reason for that is because that factory only generates one bean depending on the platform (maybe I should've called the factory `HelloFactory`). That's why specifically mentioned the `ExtendsHello`. `Hello` is the base class and is enough for the majority of the classes, but some classes need the additional context of the `ExtendsHello` (which is only used when `@BCustomScope` applies.)

[...]

PR #377, opened 2023-07-20
Add support for `@Profile` annotation - conditional wiring via profiles

SentryMan:
Adds a `@Profile` annotation for conditional wiring. Solves #304 from a certain point of view.


SentryMan:
hmm

SentryMan:
Okay I'm good with this

PR #386, opened 2023-07-27
Fix for Factory with multiple "Builder" types - short names all exactly match Builder

rbygrave:
When there are multiple Builders like A0.Builder and A1.Builder then we really need to use the full type name.

This change is to use the full type name when the short name matches Builder or Generated.


inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
```diff
@@ -418,6 +420,14 @@ void builderGetDependency(Append writer, String builderName, boolean forFactory)
       writer.append(")");
     }
 
+    private String safeShortName(String fullType) {
```
SentryMan, line 423: perhaps we could do what we did with jsonb

SentryMan: or was it http?

PR #387, opened 2023-07-27
Use Short Name for Nested Types

SentryMan:
Now generated `$DI` classes will use the more readable short name instead of the painfully long fully qualified names.

- Nested type instances will generate with enclosing type (`A.B value` instead of `some.package.A.B value`)
- Generated `_TYPE` variables will attempt to use short names if possible, but will revert to FQN when there are conflicts.


rbygrave:
I love this PR !!! :)

PR #393, opened 2023-08-27
Add option to disable automatic failure when no modules are found

SentryMan:
Now can do 
```java
BeanScope.builder().failOnEmpty(false).bean(someBean).build();
```
without an exception being thrown

### DIFF
--- a/inject/src/main/java/io/avaje/inject/BeanScopeBuilder.java
+++ b/inject/src/main/java/io/avaje/inject/BeanScopeBuilder.java
@@ -230,10 +230,14 @@ public interface BeanScopeBuilder {
    */
   BeanScopeBuilder addPostConstruct(Consumer<BeanScope> postConstructHook);
 
-  /**
-   * Add hook that will execute before this scope is destroyed.
-   */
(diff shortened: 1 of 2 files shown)


rbygrave:
What is the use case for using this?  Why will people want to turn this on?

SentryMan:
I got asked for it, and I didn't really see the harm in adding.  I suppose it's for cases where people provide beans exclusively by plugin and the `beans` method.

rob-bygrave:
We can merge this in but I suspect if this is a valid use case we don't expect to log WARNING and such a big message as that is what has been explicitly chosen. 

That is, if this is explicitly chosen I'd expect INFO level message almost saying "No modules found" because they should know all about the rest of it?

rob-bygrave:
Note that the requested use case included a supplied bean that was AutoClosable, and that would currently not work in that there is currently no check on supplied beans for AutoClosable that then adds them to the preDestroyList.

Additionally doing that would be a [maybe breaking] change in behaviour. Hmm.

SentryMan:
Not really seeing the utility, if some more arguments are provided, we can reconsider

PR #398, opened 2023-09-06
With Component.Import check for public constructor

rbygrave:
So the DI class to wire the imported component will be in the same package as the component unless it has a public constructor, and then it can be put into a
di subpackage.


SentryMan:
does this mean we can remove the new attribute I added to `@Import`?

rbygrave:
> does this mean we can remove the new attribute I added to @Import?

Well yeah I think we could actually.

SentryMan:
let's toss it then

rbygrave:
Ok yeah, makes sense cause we probably don't expect people to use it so yup - ok, I'll do that.

SentryMan (approved):
LGTM

PR #405, opened 2023-09-21
Add support for PreDestroy priority to control ordering

rbygrave:
This allows control of the ordering of the PreDestroy methods. It does this using a int priority value. The default/baseline value is 1000. Setting values less than that means those methods run earlier and values over than means those methods run later.

e.g. using `io.avaje.inject.PreDestroy` `priority` attribute:

```java

// Run before all the normal pre-destroy methods
@PreDestroy(priority = 100)
void close() {... }


// Run after all the normal pre-destroy methods
@PreDestroy(priority = 2000)
void close() {... }
```

e.g. using `@Factory @Bean` ... added the `destroyPriority` attribute. 

```java
@Factory
class MyFactory {

  @Bean(destroyMethod = "shutdown", destroyPriority = 1500)
  Hello data() {
    ...
  }
}
```

### DIFF
--- a/blackbox-test-inject/src/main/java/org/example/myapp/ExampleService.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/ExampleService.java
@@ -1,5 +1,6 @@
 package org.example.myapp;
 
+import io.avaje.inject.PreDestroy;
 import jakarta.inject.Singleton;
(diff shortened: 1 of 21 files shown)


SentryMan (approved):
LGTM

PR #407, opened 2023-10-07
Don't register beans about to be proxied

SentryMan:
This prevents cases where the proxy class is registered after it's needed.


rob-bygrave:
I think this was looking to fix the same issue as https://github.com/avaje/avaje-inject/pull/409 ?

PR #411, opened 2023-10-14
Enable Annotation processing in 22

SentryMan:
It got disabled

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,20 @@ APT-based dependency injection for server-side developers - https://avaje.io/inj
   <scope>provided</scope>
 </dependency>
 ```
+
+**JDK 22+**
+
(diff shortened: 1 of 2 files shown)


SentryMan:
I'm thinking it'll be easier, in the long run to merely update the parent rather than modifying this in all the repos.

rbygrave:
Yes, agreed. I was just happy to merge it in to A) recognise your work there + include the README change and B) double check the EA build was all cool.

I think we now do another PR that uses the parent pom v3.12.

rbygrave:
Plus if people are looking at the git history then it's nice and explicit as to what was done to get the JDK 22 build to pass. The parent pom bump is going to be less obvious to other people.

PR #431, opened 2023-10-17
Fix resilience4j test

SentryMan:


SentryMan:
hmm, it seems the problem is because resilience4j only takes java 17

Issue #439, opened 2023-10-23
[enhancement] Declarative exception handling for @Factory or @Inject


Mechite:
Simple proposal, where given a constructor in `@Singleton` or `@Component` marked class, or `@Bean` marked method in `@Factory` class:

```java
@Component
final class ExampleComponent {

    @Inject
    ExampleComponent() throws WhateverException {...}
}

@Factory
final class ExampleFactory {

    @Bean
    Example buildExample() throws WhateverException {...}
}
```

it would read the `throws` list and surround that with a try catch that simply throws a runtime exception if they are thrown. This would make it easy to make trivial issues completely fail wiring.

```java
try {
      var factory = builder.get(ExampleFactory.class);
      var bean = factory.buildExample();
      var $bean = builder.register(bean);
      builder.addPreDestroy($bean::close);
} catch (Throwable throwable) {
    throw new RuntimeException("failed to wire .....................", throwable);
}
```

PR #440, opened 2023-10-24
#439 Handle case where Factory method throws exception

rbygrave:
Adds a try catch block to the generated code


SentryMan:
Can we also handle cases where method injection and constructors throw as well?

SentryMan (approved):

SentryMan (approved):

rbygrave:
> method injection and constructors throw as well?

Ok, we have support for both of these now.

SentryMan (approved):

PR #444, opened 2023-10-25
Support RequestScope with Helidon 4.0.0 (Nima)

rbygrave:


rbygrave:
Ah yes, merging this and it will only work with Java 21 as Helidon 4.0 is Java 21+ only.

PR #467, opened 2023-12-03
Support Assisted Injection via Factory Codegen

### LINKED ISSUE #466: hope to support `@Assisted...` Annotation
I am using Guice. leveraging `@AssistedInject` and `@Assisted`, its factory module can create **new instance with parameter**.
Avaje seems to lack this feature.

Refer to: https://github.com/google/guice/wiki/AssistedInject

SentryMan:
- when a bean is annotated with `@AssistFactory` a factory class is generated which will handle all managed injections while accepting user dependencies
- supports constructor, field, and method assisted injection

Note the example below has been edited to reflect the built behaviour.

Given a factory interface with a single method:
```java
public interface CssFactory {

  Scanner scanner(String path);
}
```

A class with `@AssistFactory(CssFactory.class)`:
```java
@AssistFactory(CssFactory.class)
class CssScanner implements Scanner {

  private final Path path;
  private final MyInjectedComponent myComponent;

  CssScanner(@Assisted Path path, MyInjectedComponent myComponent) {
    this.path = path;
    this.myComponent = myComponent;
  }

  @Override
  public String scan() {
    return "scanWith|path=" + path + " | myComponent =" + myComponent.hi();
  }
}
```
An assist factory class will be generated. (Naturally, it too will be processed and have a DI class created)
```java
@Named("css")
@Component
final class CssScanner$AssistFactory implements CssFactory {

  private final MyInjectedComponent myComponent;

  CssScanner$AssistFactory(MyInjectedComponent myComponent) {
    this.myComponent = myComponent;
  }

  /**
   * Fabricates a new CssScanner.
   */
  @Override
  public CssScanner scanner(Path path) {
    var bean = new CssScanner(path, myComponent);
    return bean;
  }
}
```
Then the `CssFactory ` can be used like:.

```java
@Component
public class CssThing {

  private final CssFactory factory;

  CssThing(CssFactory factory) {
    this.factory = factory;
  }

  public String scan(Path path) {
    Scanner scanner = factory.scanner(path);
    return scanner.scan();
  }
}
```

will resolve #466


SentryMan:
Hmm, I can do better

rbygrave:
Adding tests. Noting that when there is no target factory the generated AssistedFactory can't be injected into other components (with compilation error I believe because it does not exist in the first compilation round).

Thinking that we should make the target factory required at this stage - that feels like the correct thing to do at this stage.  Later on if there is demand for not having a target factory interface then that can be revisited / looked at again.

SentryMan:
> it does not exist in the first compilation round

hmm, this highlights another potential problem in that inject doesn't know how to handle beans that directly use generated code.

PR #470, opened 2023-12-09
Use CodeGen for AOP Fallback

SentryMan:
The current way of using fallback methods is pretty obtuse, so this adds

- Adds a new annotation to mark a method as a fallback for AOP.
- will generate code to register a fallback method to an invocation
- Adds new method to `Invocation` to invoke a registered fallback method
- also will use method references in generated proxies if possible

Given stuff like this:
```java

  @MyRetry
  public String retry2(String param0, int param1) {
    retryWithFallbackCounter++;
    throw new IllegalArgumentException("Retry2Fail[" + param0 + "," + param1 + "]");
  }

  @AOPFallback("retry2")
  public String fallbackRetry2(String param0, int param1, Throwable e) {
    return "fallbackRetry2-" + param0 + ":" + param1 + ":" + e.getMessage();
  }
```
this will get generated
```java
  @Override
  public java.lang.String retry2(String param0, int param1) {
    var call = new Invocation.Call<>(() -> super.retry2(param0, param1))
      .with(this, retry20, param0, param1)
      .fallback(ex -> fallbackRetry2(param0, param1, ex));
    try {
      retry20MyRetry.invoke(call);
      return call.finalResult();
    } catch (RuntimeException ex) {
      ex.addSuppressed(new InvocationException("retry2 proxy threw exception"));
      throw ex;
    } catch (Throwable t) {
      throw new InvocationException("retry2 proxy threw exception", t);
    }
  }

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,9 +25,9 @@ APT-based dependency injection for server-side developers - https://avaje.io/inj
 </dependency>
 ```
 
-**JDK 22+**
+**JDK 23+**
 
-In JDK 22+, annotation processors are disabled by default, you will need to add a flag in maven to re-enable the processor.
+In JDK 23+, annotation processors are disabled by default, you will need to add a flag in maven to re-enable the processor.
 ```xml
 <plugin>
   <groupId>org.apache.maven.plugins</groupId>
--- a/blackbox-test-inject/src/main/java/org/example/myapp/r4j/MyExample.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/r4j/MyExample.java
@@ -3,6 +3,7 @@ package org.example.myapp.r4j;
 import org.example.myapp.resilience4j.MyRetry;
 
 import io.avaje.inject.Component;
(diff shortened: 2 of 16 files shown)


SentryMan:
@rbygrave waiting for your thoughts before I start in on the method signature validation

rbygrave:
> waiting for your thoughts before I start on the method signature validation

So yes I'm happy so far.

> start on the method signature validation

I'm not sure specifically what you mean there.  Maybe the new methods on Invocation?

SentryMan:
yeah, I'm good with this

SentryMan:
Feel free to change the names of method/class names

rbygrave:
> change the names of method/class names

For myself, I feel the ideal name for `@AOPFallback` would be `@Fallback` but we can't do that at the moment as that is the name of the deprecated interface that will be removed at some stage.

This is a really nice improvement for doing fallback methods - way better than the existing approach.

PR #477, opened 2023-12-23
Fix Lost Imports in Generated Code

### LINKED ISSUE #476: Annotation processor generate source lost import
I have an annotated  factory class
```java
@Factory
public class JMXFactory {

    @Bean
    MemoryMXBean memoryMXBean(){
        return ManagementFactory.getMemoryMXBean();
    }

    @Bean
    ClassLoadingMXBean classLoadingMXBean(){
        return ManagementFactory.getClassLoadingMXBean();
    }

    @Bean
    Executor executor(){
        return Executors.newVirtualThreadPerTaskExecutor();
    }
}
```

So I inject MemoryMXBean in my controller
```java
package org.supervisor.server.core.controller;

import io.avaje.http.api.Controller;
import io.avaje.http.api.Get;
import jakarta.inject.Inject;
import org.supervisor.server.core.dto.SystemOverviewDto;

import java.lang.management.ClassLoadingMXBean;


@Controller("/system")
public class SystemController {

    @Inject
    ClassLoadingMXBean classLoadingMXBean;

    @Get("/overview")
    public SystemOverviewDto overview() {
        return SystemOverviewDto.builder()
                .bitness(10)
                .build();
    }
}
```

And there is generate source 
```java
package org.supervisor.server.core.controller;

import io.avaje.inject.spi.Builder;
import io.avaje.inject.spi.Generated;
import org.supervisor.server.core.controller.SystemController;

/**
 * Generated source - dependency injection builder for SystemController.
 */
public final class SystemController$DI  {

  /**
   * Create and register SystemController.
   */
    if (builder.isAddBeanFor(SystemController.class)) {
      var bean = new SystemController();
      builder.addInjector(b -> {
        // field and method injection
        $bean.classLoadingMXBean = b.get(ClassLoadingMXBean.class);
      });
    }
  }
}
```
At `$bean.classLoadingMXBean = b.get(ClassLoadingMXBean.class);` will error cause the `import java.lang.management.ClassLoadingMXBean;` lost in generate source

SentryMan:
fixes #476

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/GenericType.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/GenericType.java
@@ -25,13 +25,9 @@ final class GenericType {
    */
   static String trimGenericParams(String rawType) {
     int start = rawType.indexOf('<');
-    if (start > 0) {
-      // no package for any generic parameter types
-      if (rawType.indexOf('.', start) == -1) {
-        if (rawType.lastIndexOf('>') > -1) {
-          return rawType.substring(0, start);
-        }
-      }
+    // no package for any generic parameter types
+    if (start > 0 && rawType.indexOf('.', start) == -1 && rawType.lastIndexOf('>') > -1) {
+      return rawType.substring(0, start);
     }
     return rawType;
   }
@@ -108,7 +104,9 @@ final class GenericType {
   }
 
   private static boolean includeInImports(String type) {
-    return type != null && !type.startsWith("java.lang.") && type.contains(".");
+    return type != null
+        && type.contains(".")
+        && (!type.startsWith("java.lang.") || Character.isLowerCase(type.charAt(10)));
   }
 
   /**
--- a/inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
@@ -227,7 +227,7 @@ final class TypeExtendsReader {
           .noneMatch(g -> typeElement(g.mainType()) == null);
 
       interfaceTypes.add(knownType ? rawType : GenericType.removeParameter(rawType));
-      if (!rawType.startsWith("java.lang.")) {
+      if (!rawType.startsWith("java.lang.") || Character.isLowerCase(rawType.charAt(10))) {
         for (final TypeMirror supertype : types().directSupertypes(anInterface)) {
           readInterfacesOf(supertype);
         }


SentryMan:
@rbygrave please review and deploy an RC when you get time

inject-generator/src/main/java/io/avaje/inject/generator/GenericType.java
```diff
@@ -108,7 +104,9 @@ void addImports(ImportTypeMap importTypes) {
   }
 
   private static boolean includeInImports(String type) {
-    return type != null && !type.startsWith("java.lang.") && type.contains(".");
+    return type != null
+        && type.contains(".")
+        && (!type.startsWith("java.lang.") || Character.isLowerCase(type.charAt(10)));
```
rbygrave, line 109: Not safe if length is less than 10

inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
```diff
@@ -227,7 +227,7 @@ private void readInterfacesOf(TypeMirror anInterface) {
           .noneMatch(g -> typeElement(g.mainType()) == null);
 
       interfaceTypes.add(knownType ? rawType : GenericType.removeParameter(rawType));
-      if (!rawType.startsWith("java.lang.")) {
+      if (!rawType.startsWith("java.lang.") || Character.isLowerCase(rawType.charAt(10))) {
```
rbygrave, line 230: Not safe if rawType length is less than 10

.github/workflows/build.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java_version: [11,17,21]
+        java_version: [11,17,21,22-ea]
```
rbygrave, line 15: There is the separate jdk-ea build so I don't know why we are adding 22-ea here? I don't think we want this here.

SentryMan: the EA build is currently running with JDK 23-ea (hence why it is failing)

inject-generator/src/main/java/io/avaje/inject/generator/GenericType.java
SentryMan: it is though?  that code will only trigger if the fqn starts with `java.lang`

inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
SentryMan: there's no situation where a java.lang class has less than 10 characters in the fqn

inject-generator/src/main/java/io/avaje/inject/generator/GenericType.java
SentryMan: effectively the code is
```java
  private static boolean includeInImports(String type) {
    if (type != null && type.contains(".")) {
      if (type.startsWith("java.lang.") && Character.isLowerCase(type.charAt(10))) {
        return false;
      }
      return true;
    }
    return false;
  }
```

PR #479, opened 2023-12-30
Add a switch to determine multi-module wiring order at compile time.

### LINKED ISSUE #478: Add a mechanism to determine multi-module wiring order at compile time. (for multi-module projects)
Currently, bean DI classes are created and their dependencies validated at compile time, but determining the wiring order of multi module projects is still performed at runtime. I think we should move this determination to the compile phase.

SentryMan:
So the thought here is that we can directly read the `requires` and `provides` via the external provider and determine the wiring order in the generator.

Then we can generate a class like this:

```java
public final class CompiledOrder implements ModuleOrdering {

  private final Module[] sortedModules = new Module[2];
  private static final Map<String, Integer> INDEXES =
      Map.ofEntries(
          entry("org.example.external.aspect.sub.ExampleExternalAspectModule", 0),
          entry("org.other.one.OneModule", 1));
  @Override
  public List<Module> factories() {
    return List.of(sortedModules);
  }

  @Override
  public Set<String> orderModules() {
    return INDEXES.keySet();
  }

  @Override
  public void add(Module module) {
    final var index = INDEXES.get(module.getClass().getTypeName());

    if (index != null) {
      sortedModules[index] = module;
    }
  }

  @Override
  public boolean isEmpty() {
    return sortedModules.length == 0;
  }
}
```
which knows ahead of time what goes where.

- adds a new attribute to InjectModule to toggle this mode
- generator will perform wiring checks and warn if no modules that satisfy conditions are in the class/module path
- adds new interface for determining wiring order
- adds new functions to the maven plugin to provide the module and their dependency information

should fix #478


rbygrave:
I've broken something with my last commits there trying to resolve the merge conflict.  Not sure if we want to remove/revert those ones and try again.

PR #488, opened 2024-01-24
Support qualifiers with members

SentryMan:
Can now use a qualifier's members to help distinguish beans:

So given:
```java
@Qualifier
@Retention(RetentionPolicy.RUNTIME)
public @interface TempQualifier {
  Scale[] value();

  String someOtherString();

  int defaultVal() default 0;

  NestedAnnotation[] inject() default {@NestedAnnotation()};

  enum Scale {
    CELSIUS,
    FAHRENHEIT,
  }

  @interface NestedAnnotation {

    Inject[] inject() default {};
  }
}

@Singleton
class Meters {

  Thermometer imperial;
  Thermometer metric;

  public Meters(
      @TempQualifier(value = Scale.FAHRENHEIT, someOtherString = "far") Thermometer imperial,
      @TempQualifier(value = Scale.CELSIUS, someOtherString = "celsi") Thermometer metric) {
    this.imperial = imperial;
    this.metric = metric;
  }
}
```

this will generate

```java
public final class Meters$DI  {

  /**
   * Create and register Meters.
   */
    if (builder.isAddBeanFor(Meters.class)) {
      var bean = new Meters(builder.get(Thermometer.class,"@tempqualifier(defaultval=0, inject={@nestedannotation(inject={})}, someotherstring=\"far\", value={fahrenheit})"), builder.get(Thermometer.class,"@tempqualifier(defaultval=0, inject={@nestedannotation(inject={})}, someotherstring=\"celsi\", value={celsius})"));
      builder.register(bean);
    }
  }
}
```


SentryMan:
Curious, it seems JDK 11 has some weird issue that prevents this from working.

SentryMan:
It seems that in JDK 11, the qualifier code generates differently, it'll wire, but the ability to use test mocking is impaired.

SentryMan:
the problem here seems to be that there's a discrepancy with how the annotation are rendered at runtime with `Annotation#toString()`, might have to only support a single `value` member at this time.

SentryMan:
I managed to get it working with regex, with the only limitation being that qualifiers with annotation members are only supported if the nested annotation type has only a single member.

inject-test/src/test/java/org/example/injectextension/WithExtnNamedSpyTest.java
```diff
@@ -20,7 +20,7 @@ class WithExtnNamedSpyTest {
   @Spy @Blue
   SomeStore blueStore;
 
-  @Spy @Named("green")
```
rbygrave, line 23: Why change these string qualifiers to have "@" prefix like "green" to "@green"?

SentryMan: It never made sense to me that the `@` symbol was omitted. It's an annotation qualifier so excluding `@` feels arbitrary

SentryMan: it makes less sense when we allow annotation qualifiers since then it would look like `green(value=21)` instead of `@green(value=21)`

SentryMan: I can update the PR to omit `@` again if you so desire, but I think it looks better this way

rob-bygrave: The use of `@Named()` takes a string and there is no requirement for that string to start with `@` - and so it becomes unclear as to whether there is a breaking change to this. 

The changes in the PR typically indicate what tests were broken and needed to change in order to support the change (and instead I think that isn't clear). If we want to do a _"style change to the tests"_ then that should be in a separate PR.

SentryMan: Now you only need `@` if the target bean uses a qualifier annotation, all other cases are the same

rob-bygrave:
Cool. So we probably should add a new test for the case where someone uses `@Named("@Foo")` [to be explicit that the `@` is not trimmed - anyone using that convention isn't broken by this change].

I'm also thinking there could be a test for the "simple case" where the qualifier only has 1 attribute? Like what was the motivation for this feature / I'm not expecting anyone to use nested attributes so `TempQualifier ` to me looks like it would not match the common case (it's more complex than the common case which is good for testing but to me looks like we don't actually test the commin/expected case).

SentryMan:
ehh, I reverted it, we could always do it later

SentryMan:
>  Like what was the motivation for this feature / I'm not expecting anyone to use nested attributes

Was looking at [Weld and got inspired.](https://docs.jboss.org/weld/reference/latest/en-US/html/injection.html#_qualifiers_with_members) In any case, it helps cut down  on annotation overload so I thought it was a good idea.

PR #499, opened 2024-02-12
Fix for @InjectTest with inherited field injection

rob-bygrave:
With a test class with inheritance and field injection on inherited fields, those fields were not being injected.

This fixes this issue.

### DIFF
--- a/inject-test/src/test/java/org/example/inherit/MyOneTest.java
+++ b/inject-test/src/test/java/org/example/inherit/MyOneTest.java
@@ -1,0 +1,22 @@
+package org.example.inherit;
+
+import io.avaje.inject.test.InjectExtension;
+import jakarta.inject.Inject;
+import org.example.inherit.notpublic.PubExposed;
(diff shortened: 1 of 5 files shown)


SentryMan (approved):
LGTM

inject-test/src/test/java/org/example/inherit/MyOneTest.java
```diff
@@ -0,0 +1,22 @@
...
+@ExtendWith(InjectExtension.class)
+public class MyOneTest extends MyOneAbstract {
+
+  @Inject
+  PubExposed something;
+
+  @Test
+  void test_expect_inheritedFieldIsInjected() {
+    assertThat(something).isNotNull();
+    assertThat(steamer).isNotNull();
```
rob-bygrave, line 20: This [inherited] steamer field was not getting injected

SentryMan:
I had forgotten that we could do that with abstract classes

PR #517, opened 2024-03-09
Fix Factory Providers

SentryMan:
- Fixes generated Modules not compiling in certain cases when using `Provider` in factory classes
- Fixes DI classes not compiling with Providers of Generic Types
- remove duplicate auto-provides in dependency metadata


inject-generator/src/test/java/io/avaje/inject/generator/models/valid/provider/FactoryProvider.java
```diff
@@ -0,0 +1,23 @@
+package io.avaje.inject.generator.models.valid.provider;
```
rbygrave, line 1: We need to move these classes into blackbox and add a test that asserts the behaviour.

PR #519, opened 2024-03-11
Avoid locking the lazy provide if cached

SentryMan:
as the title says

### DIFF
--- a/inject/src/main/java/io/avaje/inject/spi/LazyProvider.java
+++ b/inject/src/main/java/io/avaje/inject/spi/LazyProvider.java
@@ -4,9 +4,7 @@ import jakarta.inject.Provider;
 
 import java.util.concurrent.locks.ReentrantLock;
 
-/**
- * Wrap a Provider for lazy initialisation.
- */
+/** Wrap a Provider for lazy initialisation. */
 final class LazyProvider<T> implements Provider<T> {
 
   private final ReentrantLock lock = new ReentrantLock();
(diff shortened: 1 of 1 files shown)


rob-bygrave:
> change ProviderPromise to retrieve the provider from the bean map when get is called

Why do that? This is changing the semantics?

SentryMan:
due to certain [peculiarities with re-authenticating mongo](https://jira.mongodb.org/browse/JAVA-3896) I'm trying to register a `Provider` that will rotate the credentials and give the updated `MongoCollection`. It seems though that when I call `get` after wiring the provider I get a null pointer

rbygrave:
Well ok but I think that means we need 2 separate PRs. We need to document the NPE bug as a bug with ideally a stack trace that folks might be able to find etc. Plus we need the history as people read it via the PR titles to reasonably reflect the change (and so currently the title does not reflect anything about a NPE bug).

So yeah, I think we need to split this into 2 PRs for those reasons.

rbygrave:
> It seems though that when I call get after wiring this happens

Technically the get is being called **_during_** wiring. That is the constructor is called during wiring (and wiring has not completed yet and this is why the NPE).  I think the expectation would have been to use a `@PostConstuct` here - Hmm.

rbygrave:
My thought is that instead of `Provider<MongoCollection>` a `Supplier<MongoCollection>` could have been used here, a bit like: https://github.com/avaje/avaje-inject/pull/521/files

In this way the `Supplier.get()` method can be called inside the constructor (during wiring).

PR #522, opened 2024-03-12
Enable provider calls during wiring

SentryMan:
Augments the `ProviderPromise` to retrieve the provider when `get` is called.

### DIFF
--- a/inject-test/src/test/java/org/example/coffee/provider/FactoryProvider.java
+++ b/inject-test/src/test/java/org/example/coffee/provider/FactoryProvider.java
@@ -19,7 +19,8 @@ public class FactoryProvider {
   }
 
   @Bean
-  Provider<Supplier<String>> supply() {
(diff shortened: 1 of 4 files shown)


rbygrave:
Yeah, need to think about this ... e.g. Change the order of the methods in the `FactoryProvider` and it breaks - why? because the kind-of-point of using Provider for DI is to delay the injection until the Post-Construct phase. This means that for dependency ordering all the Provider dependencies are ignored. 

That is, the design use of Provider is to delay that actual wiring. The failing test here calls `get()` instead a `@Bean` factory method (which is similar to it being called inside a constructor) and that changes the requirement that it must exist **_during_** wiring.  If that becomes the requirement then ... the DI ordering needs to change to take Provider into account ... and then that means that it can't be used for the original task which is to resolve circular dependencies.

I realise this is a bit confusing. It is tempting to capture the NPE and throw a more meaningful error along the lines that you might actually want to use Supplier and not Provider. Hmmm.

PR #524, opened 2024-03-13
Add Version Numbers to Generated Code

SentryMan:
I mean why not?

```java
@Generated("io.avaje.inject.generator:9.12-RC3")


rob-bygrave:
I'm not keen on this at all. So yeah, you'll need to explain why you think this is a good idea.

SentryMan:
I'm not that attached to it

PR #540, opened 2024-04-13
Fix Prototype Factory Providers

SentryMan:
Before, prototype scoped Providers were generating incorrectly.

- Fixes ProtoType Providers
- Simplifies lazy inject generation (it turns out registering as a normal provider is already lazy)

Now can do:
```java
@Factory
public class FactoryProvider {

  @Bean
  @Prototype
  Provider<Integer> random() {
    return new Random()::nextInt;
  }
}
```

And this will generate: 

```java
public final class FactoryProvider$DI  {
  /**
   * Create and register Provider via factory bean method FactoryProvider#random().
   */
  public static void build_random(Builder builder) {
    if (builder.isAddBeanFor(Integer.class)) {
      var factory = builder.get(FactoryProvider.class);
      var bean = factory.random();
      builder.asPrototype().registerProvider(bean);
    }
  }
}
```


rbygrave:
> Fixes ProtoType Providers

In that a `@Prototype` `Provider<T>` registers a `Provider<Provider<T>>`  - with double nested Provider and really it should only be `Provider<T>` in this case.

rbygrave:
Thanks. I've ended up pulling this out into 2 separate PRs because I kind of needed to do that to understand the changes and see the impact and also because we should really have bug fixes in their own PR and not combined with other things like a refactor.

So yeah, created #541 and #542 and yes these are pretty much exactly what is in this PR.

Cheers, Rob.

PR #545, opened 2024-04-18
Change blackboxtest to use maven plugin extensions true

rbygrave:


SentryMan:
we might get more mileage using maven's pluginManagement in a parent pom

rbygrave:
> pluginManagement in a parent pom

parent poms are not great in my book in a inheritance versus composition way. I think this fails because the maven plugin is local in the CI build.

PR #548, opened 2024-04-20
Qualifiers preserve their case (no longer forced lower case) for BeanEntry and @QualifierMap keys (but still using case insensitive matching for wiring)

SentryMan:
Idk about this one, while I do think it's only natural that qualifiers be case sensitive, this can break a lot of stuff

----
**_Edit:_**
- So qualifier names / keys in a `@QualifierMap` `Map<String,T>` will preserve their case (no longer forced toLowerCase)
- Qualifier names returned by BeanEntry.qualifierName() also preserve their case  (no longer forced toLowerCase)

The wiring will still be case insensitive such that `@Named("red")` and `@Red` are both matched the same via equalsIgnoreCase being used internally.

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/ExternalProvider.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/ExternalProvider.java
@@ -31,16 +31,17 @@ final class ExternalProvider {
 
   private static final boolean injectAvailable = moduleCP();
   private static final Map<String, List<String>> avajePlugins = Map.ofEntries(
-      entry("io.avaje.jsonb.inject.DefaultJsonbProvider", of("io.avaje.jsonb.Jsonb")),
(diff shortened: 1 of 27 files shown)


rbygrave:
> only natural that qualifiers be case sensitive, this can break a lot of stuff

For myself I see it more that avaje-inject can keep the original case but use case-insensitive matching. I think that is the more realistic approach, with the view that it is a breaking change for `@QualifiedMap Map<String, T>` with the keys now keeping their original case (rather than all forced toLowerCase).

SentryMan:
> For myself I see it more that avaje-inject can keep the original case but use case-insensitive matching.

I'm fine with this then

rob-bygrave:
> I'm fine with this then

Cool.

I think this is an improvement in that the qualifier map keys being forced to all lower case is kind of odd or unexpected. Also preserving the case of the qualifiers gives people a better chance of doing something like an EnumMap.

PR #553, opened 2024-04-25
Adds `@External` annotation

SentryMan:
Adds a similar annotation to `@Nullable` such that compile time validations can be disabled for beans not managed by Inject, but will still fail at runtime if the bean is not present. An example use case is beans provided by Jooby

- Adds an `@External` marker annotation 
- updates generator to use this new annotation
- exempts Nullable/Optional beans from strict wiring calculus
- renames the processor


rbygrave:
> renames the processor

Any reason for that?

SentryMan:
> Any reason for that?

Aesthetics

SentryMan:
hmm wait a sec

SentryMan:
Okay, I'm good with this

ponyjoy:
@SentryMan 
```java
@Factory
public class JdbiFactory {
    private static final String JDBC_URL = "jdbc:mysql://localhost:3306/myapp?serverTimeZone=Asia/Shanghai";
    private static final String DB_USER = "demo";
    private static final String DB_PASSWORD = "demo";
    private static Jdbi JDBI = null;

    @Bean
    public DataSource getDataSource() {
        HikariConfig config = new HikariConfig();
        config.setJdbcUrl(JDBC_URL);
        config.setUsername(DB_USER);
        config.setPassword(DB_PASSWORD);
        DataSource ds = new com.zaxxer.hikari.HikariDataSource(config);
        return ds;
    }

    @Bean
    public Jdbi getJdbi(DataSource ds) {
        JDBI = Jdbi.create(ds);
        JDBI.installPlugin(new org.jdbi.v3.sqlobject.SqlObjectPlugin());
        return JDBI;
    }
}

public class JdbiTodoService implements TodoService {

    final Jdbi jdbi;

    //  @Inject  
   //need to add @External before Jdbi parameter to compile 
    public JdbiTodoService(Jdbi jdbi) {
        if (jdbi == null) {
            throw new IllegalArgumentException("jdbi is null");
        }
        this.jdbi = jdbi;
    }
}

```
When I use version 10.1, why required to use `@External` to compile when using the constructor injection method with bean returned by the `@Bean` method? Isn’t the `Jdbi` object in this code managed by avaje-inject?

SentryMan:
is it the same in inject 10.0?

ponyjoy:
> is it the same in inject 10.0? Also does the `JdbiTodoService` service exist in a separate module than the factory?
In 10.0
does not add annotations and compiles successfully.

SentryMan:
Is it possible for you to upload an example repo to GitHub so that I can reproduce?

SentryMan:
Also is it the same on inject 10.2-RC1

ponyjoy:
@SentryMan 
[jdbc-demo.zip](https://github.com/user-attachments/files/16341578/jdbc-demo.zip)
OK, it's a demo project

SentryMan:
did you have any luck with `10.2-RC1`? Also where are you seeing this error? I'm compiling with `@External` removed and 10.1 and I don't see the error when I run `mvn clean package`

ponyjoy:
> did you have any luck with `10.2-RC1`? Also where are you seeing this error? I'm compiling with `@External` removed and 10.1 and I don't see the error when I run `mvn clean package`

It’s really strange. Without adding `@External`, the compilation passed with versions 10.0, 10.1, and 10.2-RC1. It must be because I didn’t run `mvn clean compile` before.

rob-bygrave:
There were 2 regressions for "Partial Compile" starting in version `10.0-RC5` including `10`, `10.1`, `10.2-RC*`.  
The regressions were only fixed in `10.2`.

> and I don't see the error when I run mvn clean package

You don't see these bugs with a full compile - so mvn clean package should have always worked, but partial compile with any 10.x prior to 10.2 could definitely produce those errors.

I suspect that was what this issue is. You should see no issue with 10.2 with partial compilation now.

PR #554, opened 2024-04-27
Auto Generate Event Publishers

SentryMan:
Makes publishing events more CDI-like. The generator will now create a publisher class when it sees `Event<T>` injected and a generated impl doesn't already exist.

- Remove ApplicationEvent Annotation
- Enhance generator to automatically create publisher classes
- can now generate generic publishers
- similar to CDI, adding a qualifier to an `Event<T>` dependency will create and wire a publisher that by default sends qualified events


inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
```diff
@@ -119,15 +119,16 @@ String providesAspect() {
...
-      return List.of();
-    }
     if (baseTypeIsInterface) {
       return List.of(Util.unwrapProvider(baseUType));
     }
     var autoProvides = new ArrayList<>(interfaceTypes);
     autoProvides.addAll(extendsTypes);
-    autoProvides.add(Util.unwrapProvider(baseUType));
+    if (!autoProvide || !providesAspect.isEmpty()) {
+      autoProvides.remove(baseUType);
```
rbygrave, line 128: Can you explain why this change was made?  Why was the early exit removed and changed to this?

inject-generator/src/main/java/io/avaje/inject/generator/Util.java
```diff
@@ -284,7 +284,7 @@ static String commonParent(String currentTop, String aPackage) {
   static String getNamed(Element p) {
     final NamedPrism named = NamedPrism.getInstanceOn(p);
     if (named != null) {
-      return named.value();
+      return named.value().replace("\"", "\\\"");
```
rbygrave, line 287: Why is this replace of quotes here?

inject/src/main/java/io/avaje/inject/event/Event.java
```diff
@@ -37,9 +37,16 @@
 public abstract class Event<T> {
 
   protected final List<Observer<T>> observers;
+  protected final String defaultQualifier;
```
rbygrave, line 40: Should this be renamed to `qualifier`?  Seems like `defaultQualifier` isn't really a default per se?

inject-generator/src/main/java/io/avaje/inject/generator/Util.java
SentryMan: The reason is to bring it in line with the annotation qualifiers

Issue #559, opened 2024-05-03
Reduce Service Loader Invocations 


SentryMan:
Apparently calling the ServiceLoader multiple times slows things a little bit.

I'm thinking we can define a super interface for all four of our spi classes, load them all once, and use instanceof to filter things out.

PR #566, opened 2024-05-13
Make Generic Provides Changes Backwards Compatible

SentryMan:
Upon experimentation, it seems the `Class<?>` to `Type` changes we've made are backwards incompatible. Breaking every single multi-module project/plugin library based on avaje is pretty cringe. 

- Reverts changes to Module/Plugin classes and adds deprecation notices
- Creates super classes based on `Type` for the sake of compatibility (for whatever reason this works)
- renames `PropertyRequiresPlugin` 
- modifies scope builder, generator, and maven plugin to service load the new interfaces
Requires https://github.com/avaje/avaje-spi-service/pull/18 and also takes care of #559

### DIFF
--- a/blackbox-aspect/src/main/java/module-info.java
+++ b/blackbox-aspect/src/main/java/module-info.java
@@ -3,6 +3,6 @@ module blackbox.aspect {
   requires io.avaje.inject;
 
   //remove this and compilation fails
-  provides io.avaje.inject.spi.Module with org.example.external.aspect.sub.ExampleExternalAspectModule;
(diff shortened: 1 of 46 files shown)


rob-bygrave:
IMO you gotta stop including multiple changes into a single PR like this. It generally is a pain to review and becomes an all-or-nothing merge and reduces the ability to revert individual changes etc in the future.

In addition imo you have to try and honour the existing code format. Making format changes only for the sake of code format changes increases the diff and makes more work for me.

PR #567, opened 2024-05-14
Backwards Compatible Generic Modules

SentryMan:
Adds all the backwards compatibility stuff from #566 minus the SPI changes

### DIFF
--- a/inject/src/test/java/io/avaje/inject/LegacyBeanScopeBuilderTest.java
+++ b/inject/src/test/java/io/avaje/inject/LegacyBeanScopeBuilderTest.java
@@ -16,14 +16,16 @@ import org.junit.jupiter.api.Test;
 
 import io.avaje.inject.spi.AvajeModule;
 import io.avaje.inject.spi.Builder;
-import io.avaje.inject.spi.GenericType;
+import io.avaje.inject.spi.Module;
 
 @SuppressWarnings("all")
-class BeanScopeBuilderTest {
+class LegacyBeanScopeBuilderTest {
 
   @Test
   void depends_providedByParent() {
-    DBeanScopeBuilder.FactoryOrder factoryOrder = new DBeanScopeBuilder.FactoryOrder(new TDBeanScope(MyFeature.class), Collections.emptySet(), false);
+    DBeanScopeBuilder.FactoryOrder factoryOrder =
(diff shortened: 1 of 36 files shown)


inject/src/test/java/io/avaje/inject/LegacyBeanScopeBuilderTest.java
```diff
@@ -0,0 +1,338 @@
...
+  private TDModule bc(String name, Class<?>[] provides, Class<?>[] requires) {
+    return bc(name, provides, requires, new Class[0]);
+  }
+
+  private TDModule bc(
+      String name, Class<?>[] provides, Class<?>[] requires, Class<?>[] requiresPkg) {
+    return new TDModule(name, provides, requires, requiresPkg);
+  }
+
+  private static class TDModule implements AvajeModule {
```
rbygrave, line 201: My guess is that this should implement Module and not AvajeModule

rbygrave (approved):

rbygrave (changes requested):
Need to resolve why InjectPlugin exists

rbygrave:
Did you hit this issue of desiring 10.x to be backwards compatible or did someone else?  Just wondering of the pain level. Also just to say I'm kind of not loving "AvajeModule" as a name ... so I'm hoping that something better might come to me.

SentryMan:
Yeah, I noticed that all the plugins and libraries I've made at work ceased to function

SentryMan:
> Need to resolve why InjectPlugin exists

The maven plugin/generator is unable to get the type array from legacy plugins.

SentryMan:
> don't like "AvajeModule" as a name 

I was thinking of `AvajeModule` sort of like Guice has `GuiceModule`.

I got a couple of alternate options (just spitballing here):

- BeanModule
- ScopeModule
- GraphModule
- WiringModule
- DependencyModule

rob-bygrave:
> AvajeModule sort of like Guice has GuiceModule

Yeah, that makes sense. IMO the original "Module" was a poor choice by me. "InjectModule" is taken by the annotation (which seems ok) ... so I'm definately warming to AvajeModule (it's mostly used internally).

I should get to review this again today. Great work by the way !!!

PR #568, opened 2024-05-16
maven plugin: #567 Followup, Simplify ServiceLoader of InjectPlugin

rbygrave:
I don't why we have the .filter() and ::cast for the service loading here, seems unnecessary and can simplify.


rbygrave:
I'll merge this in an release the 10.0-RC4 of the maven plugin.

PR #572, opened 2024-05-22
Rename Property Plugin from PropertyRequiresPlugin to ConfigPropertyPlugin

SentryMan:
We use it for more than the `@RequiresX` annotations. Also I don't like the name.


inject-generator/pom.xml
```diff
@@ -37,12 +37,6 @@
       <scope>provided</scope>
     </dependency>
 
-    <dependency>
```
rob-bygrave, line 40: Looks unrelated?

SentryMan: Hmm, thought it came about when I pulled from master and had conflicts

rbygrave:
I'm pondering if we just go for a simple breaking change here.  As in, likely very very few people have implemented this interface (it hasn't been around for a long time etc).  So I think it might be acceptable and simple and less messy to just go with a simple rename and breaking change.  Pondering ...

SentryMan:
> I'm pondering if we just go for a simple breaking change here.

It'll break every single one of the other avaje plugins if we do a clean break. Like the Jsonb Inject plugin for example

PR #579, opened 2024-05-28
Multi-register beans with `List` Factory Methods

SentryMan:
Now can do:

```java
@Factory
public class ListFactory {

  @Bean
  List<String> test() {
    return List.of("test1", "test2");
  }

  @Bean
  String test3() {
    return "test3";
  }
}
```

and get `"test1", "test2","test3"` when injecting a `List<String>`. (Even Spring cannot do this currently)

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
@@ -244,9 +244,14 @@ final class MethodReader {
       }
       String indent = optionalType ? "        " : "      ";
       writer.indent(indent);
+      var hasLifecycleMethods = hasLifecycleMethods();
 
-      if(multiRegister) {
-       writer.append("bean.forEach(");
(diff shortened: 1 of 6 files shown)


inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
```diff
@@ -229,9 +244,11 @@ void builderBuildAddBean(Append writer) {
       }
       String indent = optionalType ? "        " : "      ";
       writer.indent(indent);
-      if (hasLifecycleMethods()) {
-        writer.append("var $bean = ");
```
rob-bygrave, line 233: We have to keep using the returned `$bean` because with testing and using mockito the `$bean` instance can be different to `bean`. 

So this change breaks that.

SentryMan: I see, I'll probably add a test for that later.

PR #587, opened 2024-06-04
Restore autoProvides to exclude http controllers and generated http routes

rbygrave:
The change to autoProvides for Event support also meant that http controllers and http routes started to get included into autoProvides. The problem there is that http controllers are not expected to be public exported types and autoProvides with GenericType doesn't check is the type parameters are public.


SentryMan:
>  The problem there is that http controllers are not expected to be public exported types

You have all kinds of applications, a week ago we had an issue in the discord. Routes provided in one module were not picked up in the other module that gathers all the routes and starts the server.

rbygrave:
That isn't related to autoProvides though.

SentryMan:
> That isn't related to autoProvides though.

With this change if we want to have an `api` module which created all the routes, and 
a `svc` module that consumes all route beans we need to add an `@InjectModule(requires)`

PR #588, opened 2024-06-04
Restore AutoProvides to include http common interfaces

SentryMan:
From what I've seen from doing support, it's not uncommon to provide HTTP routes to other modules.

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/Constants.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/Constants.java
@@ -20,6 +20,7 @@ final class Constants {
 
   static final String PATH = "io.avaje.http.api.Path";
   static final String CONTROLLER = "io.avaje.http.api.Controller";
+  static final String HTTP_GENERATED = "io.avaje.http.api.Generated";
 
   static final String AT_SINGLETON = "@Singleton";
   static final String AT_PROXY = "@Proxy";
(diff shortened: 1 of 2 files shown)


rbygrave:
>  uncommon to provide HTTP routes to other modules.

It is common but that does not require autoProvides to work. This change breaks http apps by requiring the controllers to be public which they are not required to be (the routes are public but NEITHER need autoProvides because other modules do not have a compile time tight dependency to either controller or route.

When the route is registered via the common public interface that is how all the routes are found and registered with the http server and that does NOT require autoProvides.

If there was a problem there is was unrelated to autoProvides.

SentryMan:
`do not have a compile time tight dependency to either controller or route.`

They may not need the route class itself, but they do need the `AvajeJavalinPlugin` or `HttpFeature` to be autoprovided.

SentryMan:
> When the route is registered via the common public interface that is how all the routes are found and registered with the http server and that does NOT require autoProvides.

Not sure I follow, say I have module `A` which has the generated routes. How does module `B` know that `A` has any generated routes without putting a `@InjectModule(requires=HttpFeature.class)` on `B`?

rbygrave:
> How does module B know that A has any generated routes without putting a @InjectModule(requires=HttpFeature.class) on B?

Does module `B` have routes? I think this is the issue ... If `B` has routes then if does not matter, if `B` has _no routes_ but wants to wire/register them then yes it will need something like `@InjectModule(requires=HttpFeature.class)` or for the routes to have autoProvides.

SentryMan:
let's compromise then, I'll change this such that it will auto-provide only the HTTP common interfaces and not the route classes themselves

rbygrave:
Replaced by https://github.com/avaje/avaje-inject/pull/590

PR #591, opened 2024-06-05
Let SPI Service take care of module validation

SentryMan:
As the title says

Relies on https://github.com/avaje/avaje-spi-service/pull/22


inject-generator/src/main/java/io/avaje/inject/generator/TypeExtendsReader.java
```diff
@@ -21,7 +21,6 @@ final class TypeExtendsReader {
 
   private static final Set<String> ROUTE_TYPES = Set.of(
     "io.avaje.http.api.AvajeJavalinPlugin",
-    "io.javalin.Javalin",
```
rbygrave, line 24: What is this removal in here for?

rbygrave:
Merged via https://github.com/avaje/avaje-inject/pull/593

PR #600, opened 2024-06-10
Split Events into their own module

SentryMan:
- changes `event` package to `events`
- moves everything into its own module and adds as a transitive dep to inject
- changes generation to directly register an `Observer`


inject-events/pom.xml
```diff
@@ -0,0 +1,32 @@
...
+    <version>10.0-RC6</version>
+  </parent>
+  <artifactId>avaje-inject-events</artifactId>
+  <name>avaje inject events</name>
+  <description>avaje-inject events plugins for event publishing/consuming</description>
+  <dependencies>
+    <dependency>
+      <groupId>io.avaje</groupId>
+      <artifactId>avaje-inject</artifactId>
+      <version>10.0-RC5</version>
```
rbygrave, line 15: Note to check this after merge, probably should be RC6 and not RC5 here.

rbygrave (approved):

PR #606, opened 2024-06-10
Have the aop and events modules as opt-in dependencies rather than opt-out

rbygrave:
Currently people not wanting the avaje-inject-aop or avaje-inject-events dependencies need to exclude them.

This is the alternative where instead projects need to explicitly include those modules. An extension to this would be to rename avaje-inject to be avaje-inject-core and introduce a new avaje-inject which is just a composite of avaje-inject-core, avaje-inject-aop and avaje-inject-events.

Hmmm. Pondering this.

### DIFF
--- a/blackbox-test-inject/pom.xml
+++ b/blackbox-test-inject/pom.xml
@@ -50,6 +50,18 @@
     </dependency>
 
     <dependency>
+      <groupId>io.avaje</groupId>
+      <artifactId>avaje-inject-aop</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+
+    <dependency>
+      <groupId>io.avaje</groupId>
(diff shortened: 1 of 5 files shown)


SentryMan:
I don't mind it, we just need to update the docs

SentryMan (approved):

rbygrave:
Yeah, I'm leaning towards not doing this with the thinking that it keeps it simpler for most? people - and harder for the smaller group that want to optimise and exclude those dependencies.

An alternative is to create a avaje-inject-core (avaje-inject-minimal?) which is avaje-inject with the maven exclusions for aop and events (so we do the exclusion for them and provide that as a dependency). Could do that later based on demand/need.

PR #613, opened 2024-06-12
Refactor rename AvajeModule interface to InjectModule

rbygrave:
Note that I think this is reasonable in terms of @InjectModule being the same name in that the IDE auto-completion when using the annotation knows it's an annotation. So I don't think we get problematic use like this.

In the generated code is the one place that we do get the clash using both the annotation and interface and in this case we fully qualify the interface in the generated code.


SentryMan:
Can't say I'm fan of classes sharing the same name.

rbygrave:
Yeah, lets not merge this.

PR #616, opened 2024-06-20
Write SPIs to META-INF/generated-services so that they can be merged later via avaje-spi-service

SentryMan:
Now relies on spi-service to write spis

relies on https://github.com/avaje/avaje-spi-service/pull/25

---
Note: This ultimately is required because with the change to a common service interface of InjectExtension we now more likely get multiple services registering for the same InjectExtension api interface - so those need to be merged into the same META-INF/services file via avaje-spi-service.

### DIFF
--- a/inject-generator/src/test/java/io/avaje/inject/generator/InjectProcessorTest.java
+++ b/inject-generator/src/test/java/io/avaje/inject/generator/InjectProcessorTest.java
@@ -26,6 +26,8 @@ import javax.tools.ToolProvider;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 
+import io.avaje.spi.internal.ServiceProcessor;
+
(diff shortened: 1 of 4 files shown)


inject-generator/src/test/java/io/avaje/inject/generator/InjectProcessorTest.java
```diff
@@ -67,7 +68,7 @@ void testGeneration() throws Exception {
             List.of("--release=" + Integer.getInteger("java.specification.version")),
             null,
             files);
-    task.setProcessors(Arrays.asList(new InjectProcessor()));
+    task.setProcessors(List.of(new InjectProcessor(), new ServiceProcessor()));
```
SentryMan, line 71: @rbygrave you can probably use your IDE here to debug and understand what's going on if it's still not making intuitive sense.

PR #621, opened 2024-06-24
Use Avaje Prisms for module-validation

SentryMan:
Uses Avaje Prisms to validate modules

- removes avaje-spi's influence over spi generation
- uses avaje prisms to validate modules


inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
```diff
@@ -188,7 +198,28 @@ static void processingOver(boolean over) {
...
+                .toURL()
+                .openStream();
+        final var buffer = new BufferedReader(new InputStreamReader(file)); ) {
+
+      String line;
+      while ((line = buffer.readLine()) != null) {
+        line.replaceAll("\\s", "")
+            .replace(",", "\n")
+            .lines()
+            .forEach(ProcessingContext::addInjectSPI);
```
rob-bygrave, line 215: We need to read and register these because at line 225 that is a write **_replace_** and not a write _**append**_ ?

SentryMan: yes

PR #622, opened 2024-06-24
Adds new SPI Annotation

SentryMan:
Trying to auto-register services with avaje-spi is a dead end. So I suppose we need to make spi annotation for the generators.

### DIFF
--- a/blackbox-test-inject/src/main/java/org/example/myapp/ConfigPropertiesPlugin.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/ConfigPropertiesPlugin.java
@@ -3,10 +3,13 @@ package org.example.myapp;
 import io.avaje.config.Config;
 import io.avaje.inject.Component;
 import io.avaje.inject.spi.ConfigPropertyPlugin;
+import io.avaje.inject.spi.InjectSpi;
+
 import org.other.one.OtherComponent2;
 
 import java.util.Optional;
 
+@InjectSpi
 @Component.Import(value = OtherComponent2.class)
 public class ConfigPropertiesPlugin implements ConfigPropertyPlugin {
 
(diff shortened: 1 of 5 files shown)


SentryMan:
Feel free to change the names of the annotations to something else

rob-bygrave:
> Trying to auto-register services with avaje-spi is a dead end.

Why? What has gone wrong?

SentryMan:
> > Trying to auto-register services with avaje-spi is a dead end.
> 
> Why? What has gone wrong?

I thought it was good too, but it has a fatal flaw in that if multiple annotation processors try to use this, not all of the processor's spi interfaces will get registered. The reason is that the spi processor must be the last processor to execute after processing is over.

Trying to wrangle the apt order causes more work, not less.

rob-bygrave:
> So I suppose we need to make spi annotation for the generators.

Or not? In that there are several options to consider now right?

What we know is that we need the inject-generator AP to write it's META-INF/services and not delegate that to avaje-spi-service (due to unknown ordering of the APs). Now if inject-generator AP **_appends_** to META-INF/services (and does not override) then this allows traditional manual registration of any InjectExtension. Does that not work?

SentryMan:
> Does that not work?

that's what #621 is for. This one is because I don't like manually writing services

rbygrave:
> that's what https://github.com/avaje/avaje-inject/pull/621 is for.

Yes, got that now. I saw the emails and saw the _"Trying to auto-register services with avaje-spi is a dead end."_ ... so I came here first (so yeah I missed the previous part of the story there).

SentryMan:
hmm, an alternate strategy would be to annotation process the avaje-spi `@ServiceProvider` and modify avaje-spi-service such that it'll abort writing any of the spis claimed by our generators

rob-bygrave:
It is generally the case for avaje-inject that the only folks registering a service will be InjectPlugin authors right?  Like there might be better cases like avaje-validation but its not got good bang for buck here for avaje-inject as I see it.

SentryMan:
> It is generally the case for avaje-inject that the only folks registering a service will be InjectPlugin authors right? 

Even so, writing services is tedious. In any case, don't forget those who may want to use the Property Plugin spi or fiddle with custom module ordering logic.

PR #624, opened 2024-06-25
Process Avaje SPI @ServiceProvider for writing service META-INF/services file entry

SentryMan:
Developers looking to extend avaje-inject via one of InjectExtension interfaces can register the service implementation manually via `src/main/resources/META-INF/services/io.avaje.inject.spi.InjectExtension`.

With this change, alternatively the Developer can use `avaje-spi-service` and annotate the service implementation with `@ServiceProvider` - then the entry will be automatically created in the `target/classes/META-INF/services/io.avaje.inject.spi.InjectExtension`. The avaje-inject-generator will detect the extension and add it to the services that it needs to register.

-----
A potential alternative to #622 wherein inject will process the avaje-spi annotations instead of creating a new one.

avaje spi would also help validate that the structure of the service itself is good (public no-args constructor and the like)

relies on https://github.com/avaje/avaje-spi-service/pull/27


inject-generator/src/main/java/io/avaje/inject/generator/InjectProcessor.java
```diff
@@ -332,4 +334,11 @@ private void readInjectModule(RoundEnvironment roundEnv) {
         }
       });
   }
+
+  private void registerSPI(Set<? extends Element> beans) {
+    ElementFilter.typesIn(beans).stream()
+        .map(TypeElement::getQualifiedName)
```
rbygrave, line 340: It looks like this needs to filter on types that implement InjectExtension?

blackbox-test-inject/src/main/java/org/example/myapp/ConfigPropertiesPlugin.java
```diff
@@ -3,10 +3,13 @@
 import io.avaje.config.Config;
 import io.avaje.inject.Component;
 import io.avaje.inject.spi.ConfigPropertyPlugin;
+import io.avaje.spi.ServiceProvider;
+
 import org.other.one.OtherComponent2;
 
 import java.util.Optional;
 
+@ServiceProvider
```
rbygrave, line 12: Technically I do not think this is required

inject-generator/src/main/java/io/avaje/inject/generator/InjectProcessor.java
rbygrave: I have added a filter such that InjectProcessor only adds services that are InjectExtension ones.  Other services are expected to be handled by avaje-spi-service (or something else like avaje-jsonb or avaje-validator).

rbygrave:
@SentryMan I'll merge this in and we can adjust as necessary if any of my changes were not correct.

PR #627, opened 2024-06-25
Remove @Service from InjectExtension as not required

rbygrave:
I think it is clearer if @Service is not here as it isn't used or required.

### DIFF
--- a/inject/pom.xml
+++ b/inject/pom.xml
@@ -44,13 +44,6 @@
     </dependency>
 
     <dependency>
-      <groupId>io.avaje</groupId>
-      <artifactId>avaje-spi-service</artifactId>
-      <version>1.11</version>
-      <optional>true</optional>
-    </dependency>
-
-    <dependency>
       <groupId>org.mockito</groupId>
       <artifactId>mockito-core</artifactId>
       <version>5.12.0</version>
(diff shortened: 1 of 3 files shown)


SentryMan:
No it's needed

rbygrave:
> avaje spi only know how to avoid colliding 

It avoids collision via the hard coded exclusions, so in that way it isn't needed.  I'm thinking you agree as you reopened this PR?

SentryMan:
I closed it as a mistake before we could talk it out

rbygrave:
Ok, we can close this.  Copying description from jsonb here:


--------
This is required in order to support the use of avaje-spi-serivce `@ServiceProvider` with
other custom InjectExtension service implementations.

Custom InjectExtension implementations can be registered explicitly with META-INF/services
entries OR we can use `@ServiceProvider` for that to automatically happen.

The reason the `@Service` is required on the InjectExtension interface is so that avaje-spi-service
actually ignores it correctly (and thus allowing avaje-jsonb-generator to process it with the
generated InjectExtension implementations). Without this  avaje-spi-service incorrectly
assumes that the top level interface is the service interface and then doesn't ignore it and instead
generates an incorrect file in target/classes/META-INF/services.

PR #631, opened 2024-06-27
Make Annotations Source Retention

SentryMan:
Change annotations to have Class retention with the exception of:
- `@Component`
- `@PreDestroy`
- `@PostConstruct`

The reason for these exceptions is that for APM (Application Performance Monitoring) tools like `avaje-metrics` detect and enhance with timing all public methods of all `@Component` / `@Singleton`.  Lifecycle methods annotated with `@PreDestroy`/ `@PostConstruct` typically get excluded by default from APM so we need to detect those as well.

Note that if projects need some of these annotations to have RUNTIME or CLASS retention then please log an issue with the use case that you need to support - thanks.

### DIFF
--- a/inject/src/main/java/io/avaje/inject/AssistFactory.java
+++ b/inject/src/main/java/io/avaje/inject/AssistFactory.java
@@ -54,7 +54,7 @@ import java.lang.annotation.Target;
  * }</pre>
  */
 @Target(ElementType.TYPE)
-@Retention(RetentionPolicy.RUNTIME)
+@Retention(RetentionPolicy.SOURCE)
 public @interface AssistFactory {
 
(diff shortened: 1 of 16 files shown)


rob-bygrave:
Some of these annotations are used for APM - Application Performance Monitoring. For example, they are detected by `avaje-metrics` to enhance which can timing metrics to all public methods of all `@Component`, `@Singleton` etc and typically excluding `@PostConstruct`.  So no, we can't make them all SOURCE retention.

rbygrave:
We can re-open this, I can go through and change the ones we want to keep at RUNTIME. I'll reopen for now.

rob-bygrave (approved):

PR #645, opened 2024-07-14
Validate Qualifiers at compile time

### LINKED ISSUE #644: Dependency information for qualified beans does not include qualified name, resulting in incorrect circular dependency detection
Given the following code:
```java
// test/IncorrectCircularDependency.java
package test;

import io.avaje.inject.Bean;
import io.avaje.inject.Factory;
import jakarta.inject.Named;
import jakarta.inject.Singleton;

@Singleton
@Factory
public class IncorrectCircularDependency {
    @Bean
    @Named("parent") // for clarity, not necessary
    public String parent(@Named("child") String child) {
        throw new AssertionError("Method body unimportant");
    }

    @Bean
    @Named("child")
    public String child() {
        throw new AssertionError("Method body unimportant");
    }
}
```
```java
// test/Test.java
package test;

import jakarta.inject.Inject;
import jakarta.inject.Singleton;

@Singleton
public class Test {
    private final String parent;

    @Inject
    public Test(@Named("parent") String parent) {
        this.parent = parent;
    }

    public String getParent() {
        return parent;
    }
}
```

Despite the clearly non-circular dependency from `parent` to `child`, due to the dependency of `parent` incorrectly not using the `child` qualifier and depending on itself, avaje-inject spits out this error:
```
error: Circular dependencies detected with beans [java.lang.String:parent, test.Test]  To handle circular dependencies consider using field injection rather than constructor injection on one of the dependencies. 
   See https://avaje.io/inject/#circular
error: Circular dependency - java.lang.String:parent dependsOn java.lang.String:parent for java.lang.String
error: Circular dependency - test.Test dependsOn java.lang.String:parent for java.lang.String
```

This should not be an error and should compile cleanly.

SentryMan:
It seems previously we only validated that all the bean _types_ were in the scope at compile time without looking at the qualifiers.

- Change generated code to append qualifiers to dependency metadata
- now will confirm that all qualified beans are in the scope when compiling

resolves #644


rob-bygrave:
We should try to keep a PR focused on it's change only and not include other unrelated changes into it.  My instinct is that "fix event publishers not getting detected" and "refactor reading external modules" should have been in separate PRs?

One aspect of this is that the PR Title is important documentation, and the other is from a git history perspective we want a commit to be focused on it's change. Putting multiple things into a commit makes life harder for our future selves when we try to understand the interaction between features by looking at commit history.

SentryMan:
> My instinct is that "fix event publishers not getting detected" and "refactor reading external modules" should have been in separate PRs?

sure I can break the external one into it's own thing, but the event publisher bug only arises from the changes in this PR

rob-bygrave:
> but the event publisher bug only arises from the changes in this PR

Ok cool. Then lets keep that in as part of this PR.

PR #650, opened 2024-07-18
Support Jspecify Annotations

### LINKED ISSUE #649: Replace Avaje Lang with Jspecify?
It seems [Jspecify has released `1.0.0`](https://github.com/jspecify/jspecify/releases/tag/v1.0.0), I'm thinking we should replace all of our custom nullable annotations with the ones from jspecify.

SentryMan:
It seems due to their nature as `TYPE_USE` annotations we had not been detecting them correctly.

- replaces avaje lang with jspecify
- fixes generator to handle type use nullable annotations
- make jakarta inject `static transitive` (it's needed for compilation only)

Refer to https://jspecify.dev/docs/using#if-your-code-already-uses-jsr-305-annotations for differences we might see with this change

Also refer to https://jspecify.dev/docs/tool-conformance ... for more details on tooling support

resolves #649

### DIFF
--- a/inject/src/main/java/io/avaje/inject/spi/DBeanScope.java
+++ b/inject/src/main/java/io/avaje/inject/spi/DBeanScope.java
@@ -182,11 +182,10 @@ final class DBeanScope implements BeanScope {
   static <T> List<T> combine(List<T> values, List<T> parentValues) {
     if (values.isEmpty()) {
       return parentValues;
+    } else if (parentValues.isEmpty()) {
+      return values;
     } else {
-      if (parentValues.isEmpty()) {
-      } else {
-        values.addAll(parentValues);
-      }
+      values.addAll(parentValues);
       return values;
     }
   }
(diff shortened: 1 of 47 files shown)


inject/src/main/java/io/avaje/inject/spi/DBeanScope.java
```diff
@@ -175,10 +182,11 @@ private <T> List<T> listOf(Type type) {
   static <T> List<T> combine(List<T> values, List<T> parentValues) {
     if (values.isEmpty()) {
       return parentValues;
-    } else if (parentValues.isEmpty()) {
-      return values;
     } else {
-      values.addAll(parentValues);
+      if (parentValues.isEmpty()) {
```
rob-bygrave, line 186: This (A) creates an empty if block which I really don't like and (B) isn't related to JSpecify so really shouldn't be in this PR?

PR #659, opened 2024-07-31
Fix cases where test beans fail module validation

SentryMan:
Can't seem to replicate the issue in a test, but when I install into my local maven repo it resolves the avaje-http build issue.


SentryMan:
@rbygrave please review and deploy when you get time

PR #660, opened 2024-07-31
Generate Value Classes When on Valhalla

SentryMan:
Was testing with valhalla, and thought why not?

- will generate value classes if on a valhalla JDK build and previews are enabled
- Modifies generated module classes to work as value types (no mutable fields)

```java
@Generated
@InjectModule
public final value class ApiModule implements AvajeModule {
  @Override
  public void build(Builder builder) {
    // builder is now passed into the DI methods instead of being a field
    build_config_HttpClientFactory(builder);
  }
...
```


SentryMan:
Alright, I'm good with this

PR #665, opened 2024-08-18
Add `@BeanTypes` Annotation to Limit Injection Types

SentryMan:
- Creates `@BeanTypes` annotation to define the injectable types associated with a bean. 
- the values provided to the annotation will also override auto-provides generation

Given the class:

```java
@Component
@BeanTypes(SomeInterface2.class)
public class LimitedOtherComponent extends AbstractSuperClass
    implements SomeInterface, SomeInterface2 {}
```

the following will be generated: 

```java
public final class LimitedOtherComponent$DI  {

  /**
   * Create and register LimitedOtherComponent.
   */
    if (builder.isAddBeanFor(SomeInterface2.class)) {
      var bean = new LimitedOtherComponent();
      builder.register(bean);
    }
  }
}
```

### DIFF
--- a/blackbox-test-inject/src/main/java/org/example/myapp/beantypes/AbstractSuperClass.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/beantypes/AbstractSuperClass.java
@@ -1,0 +1,3 @@
+package org.example.myapp.beantypes;
+
+public abstract class AbstractSuperClass {}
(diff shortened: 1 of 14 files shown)


rbygrave:
There are no blackbox tests yet. So it compiles but I don't see tests in blackbox that assert the behaviour. Can we add those tests?

SentryMan:
Ran out of time unfortunately, on a road trip without anything more complex than a phone

SentryMan:
Added the tests, please review when you get time

SentryMan:
@rbygrave please deploy an RC when you find the time

PR #670, opened 2024-08-26
Breaking change - Remove support for Deprecated Interfaces (please upgrade dependencies)

SentryMan:
The generator already prevents code from generating with the old interfaces, this seals the deal.


rbygrave:
> The generator already prevents code from 

**_BUT_** currently older plugin dependencies do still work.  This PR means that they will no longer work ... with the implication that it FORCES people to update their dependencies of the various plugins.  As such it is breaking change.

Is that not the case?

SentryMan:
We can give more time then if you don't think people have upgraded their dependencies, I made the PR because I recently migrated all my job's avaje plugins to the new interfaces

rbygrave:
> We can give more time then if you don't think people have upgraded their dependencies

Yes, I think we need to give people more time.

As I see it the _ideal_ ... is for breaking changes to be in their own release with no other changes (no other bug fixes, no other PRs). I think that helps anyone who is trying to follow what is going on when a given release _ONLY_ contains breaking changes - there are no other changes included that bury/hide the important detail that the release will be a breaking change for some apps.

In addition, ideally the title of the PR is clear and obvious that it's a breaking change. I've adjusted this title to hopefully be more clear in terms that it could breaking some apps and what people are expected to do (update the dependencies).

The saying I have is "It does not hurt to be bloody obvious". People are busy, they only scan release notes at best ... the PR and Issue titles are super important and we should try and make them "bloody obvious" if we can etc.

rbygrave (approved):

PR #684, opened 2024-08-28
Add support for building against Valhalla EA

rbygrave:
Not using many `value` classes here in avaje-inject but I think this might be a decent approach/pattern if we want to build and test using Valhalla value classes. If this is a good idea/approach then could look at doing something similar with avaje-jsonb?

The guts is that there is a `/*value*/ class` placeholder that can regex replaced to `value class` (in a similar manor to the jakarta to javax approach).

### DIFF
--- a/.github/workflows/valhalla-ea.yml
+++ b/.github/workflows/valhalla-ea.yml
@@ -5,8 +5,8 @@ on:
   workflow_dispatch:
 #  push:
 #  pull_request:
-#  schedule:
-#    - cron: '39 1 * * 1,3,5'
+  schedule:
+    - cron: '39 1 * * 1,3,5'
 
 jobs:
   build:


inject-generator/src/test/java/io/avaje/inject/generator/InjectProcessorTest.java
```diff
@@ -41,6 +41,7 @@ void deleteGeneratedFiles() {
     }
   }
 
+  @Disabled
```
rbygrave, line 44: The module import gets enabled ... and that does not work against Valhalla EA. So maybe thinking of not using that module import feature.

SentryMan: can we have this test disabled only for valhalla?

rob-bygrave: Yeah maybe. The origin issue as I see it is the "module import" feature ... it gets turned on due to --enable-preview but it isn't in the Valhalla EA build so that fails. So for me it makes more sense to disable that when using Valhalla or just disable it for now until the Valhalla EA also has that feature included.

You don't perfer that idea?

Issue #692, opened 2024-10-04
Global lazy


re-thc:
Hi, are there configuration options for inject, e.g. to apply lazy globally? Spring had an option to globally mark everything as lazy.

Trying to port an application to use this, but seems tedious to maintain `@Lazy` everywhere. Thanks!

SentryMan:
Not at the moment, though that's something we can try to add

rbygrave:
> to apply lazy globally?

What is the use case? Why are you looking to apply this globally like that is a good thing to do?

I'm coming from the perspective that ideally Lazy isn't used much (Eager is preferred). In the server-side K8s case, when we start a service and respond OK to readiness we really want the service to be "Fully honest" along the lines "Yes I have started, I have initialised everything I need and I am ready to receive production traffic".  In the everything is lazy case, we are really susceptible to saying OK to readiness without having initialised much at all ... and failing when trying to handle production traffic - in my view this is not cool relative to the "fail fast" / "Don't say OK to readiness unless are all truely all ready to serve production traffic".

Lazy was initially added for the JavaFX UI case, to delay initialisation of things that were relatively expensive and could be reasonably deferred to initialise lazily. What is the use case you are looking at? Why do you think global lazy is a good idea for that use case?

re-thc:
Thanks for considering. The production case makes perfect sense.

The use case is more for dev / testing. In this case faster startup / restarts would be preferred.

Spring devtools for example does provide hot reload with custom classloader so not everything restarts. It does have its quirks though.

At the moment I'm using inject and just restarting, which can take a while with lots of services and connections.

Hoping to also create preview builds per PR and having everything lazy would speed things up.

rbygrave:
> At the moment I'm using inject and just restarting

It sounds like the case of building a Web UI (probably with server side rendering) and hence restarting the server frequently.

> to apply lazy globally?

We can't do this **_transparently_** without using a more advanced AOP approach [along the lines of a javaagent]. That is, on the side when a component is injected we are effectively look to inject a `Provider<T>` rather than a `T` ... to enable the lazy use of `T`. We can't do that without a more more advanced AOP approach.

SentryMan:
> We can't do this transparently

why not just add an annotation that changes the default behavior to make all beans lazy?

rbygrave:
> why not just add an annotation that changes the default generation behavior to make all beans lazy?

[...]

PR #698, opened 2024-10-14
Make `@Lazy` work on a package/module level

SentryMan:
Now can add `@Lazy` to a module/package to make all beans within lazy, instead of needing to annotate each bean manually.

It's one way to solve #692


SentryMan:
doesn't actually solve the problem

PR #708, opened 2024-10-24
Fix Inject-test such that it does NOT have a dependency on inject-events

SentryMan:
Currently, when inject-test is built, it has a `META-INF/services/InjectExtension` with an entry for `io.avaje.inject.events.spi.ObserverManagerPlugin`. 

This means that it has hard dependency on inject-events. When events is not present this exception occurs:

```
java.util.ServiceConfigurationError: ServiceConfiguration io.avaje.inject.spi.InjectExtension: Provider io.avaje.inject.events.spi.ObserverManagerPlugin not found
```
this pr makes it so that entry is not added to inject-test

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
@@ -205,12 +205,15 @@ final class ProcessingContext {
   }
 
   static void writeSPIServicesFile() {
-
     Optional.ofNullable(APContext.getProjectModuleElement())
         .filter(m -> "io.avaje.inject.test".equals(m.getQualifiedName().toString()))
         .ifPresent(m -> CTX.get().spiServices.remove(EVENTS_SPI));
 
     readExistingMetaInfServices();
+    if (CTX.get().spiServices.isEmpty()) {
+      // no services to register
+      return;
+    }
     try {
       FileObject jfo = createMetaInfWriterFor(Constants.META_INF_SPI);
       if (jfo != null) {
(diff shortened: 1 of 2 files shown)


rbygrave:
> It seems

We can be explicit here, that helps anyone who hits this issue and is trying to understand it. 

Currently, when inject-test is built, it is has a `META-INF/services/InjectExtension` with an entry for `io.avaje.inject.events.spi.ObserverManagerPlugin`. This means that it has this implicit dependency on inject-events.

The fix for this issue is to NOT generate that `META-INF/services/InjectExtension` at all when inject-test is built.

inject-generator/src/main/java/io/avaje/inject/generator/ProcessingContext.java
```diff
@@ -205,7 +205,15 @@ static void processingOver(boolean over) {
   }
 
   static void writeSPIServicesFile() {
+    Optional.ofNullable(APContext.getProjectModuleElement())
+        .filter(m -> "io.avaje.inject.test".equals(m.getQualifiedName().toString()))
+        .ifPresent(m -> CTX.get().spiServices.remove(EVENTS_SPI));
+
     readExistingMetaInfServices();
+    if (CTX.get().spiServices.isEmpty()) {
```
rbygrave, line 213: Are you happy with this change to only conditionally generate the services file @SentryMan ?

If so I think we can merge.

SentryMan: Yeah I'm good with this

rbygrave (approved):

PR #714, opened 2024-10-30
Change MetaData to be Comparable and sort to deterministic order

rbygrave:


SentryMan (approved):
yeah that's better than what I had

PR #718, opened 2024-10-30
Use Module Path to Find External Modules

### LINKED ISSUE #716: Use `Elements#getAllModuleElements` to read external modules
I just had a brainwave, what if we do something like:

```java
      var types = APContext.types();
      var spi = APContext.typeElement("io.avaje.inject.spi.AvajeModule").asType();
      APContext.elements().getAllModuleElements().stream()
          .filter(s -> !s.getQualifiedName().toString().startsWith("java"))
          .filter(s -> !s.getQualifiedName().toString().startsWith("jdk"))
          .filter(s -> s != APContext.getProjectModuleElement())
          .flatMap(m -> m.getEnclosedElements().stream())
          .flatMap(m -> m.getEnclosedElements().stream())
          .map(TypeElement.class::cast)
          .filter(m -> m.getKind() == ElementKind.CLASS)
          .filter(m -> m.getModifiers().contains(Modifier.PUBLIC))
          .filter(m -> m.getInterfaces().stream().anyMatch(t -> types.isAssignable(t, spi)))
          .forEach(s -> {
             // read DependencyMeta Annotations
          });
```

In this way we can read available external avaje modules without needing a maven/gradle plugin. (`InjectPlugin`s will still need a plugin though)

SentryMan:
This works but it has a few caveats:

- it will only work if the application being compiled has a `module-info.java`
- The processor can only read external `AvajeModule`s if the package is exported packages (or if the external lib does not have a `module-info.java`) 

Resolves #716


SentryMan:
@rbygrave yeah I'm good with this

rob-bygrave (approved):

PR #720, opened 2024-10-31
Remove `exports` limitation from Module Path Detection

SentryMan:
Removes the main weakness of #718 by reading the `provides` directives of all external modules.


SentryMan:
ok in this situation we need to export the `org.example.external.aspect.sub` package because some classes are used in the tests. If you look at a [failure log of a previous run,](https://github.com/avaje/avaje-inject/actions/runs/11604964049/job/32314566282) you will see that it compiled successfully and failed only on test compilation

rbygrave (approved):

PR #721, opened 2024-10-31
Allow ordering by Comparable compareTo

mintdaniel42:
We already have listByPriority and listByAnnotation but now we are able to sort it when our interface implements Comparable<T> without using any reflection or Runtime annotations

### DIFF
--- a/inject/src/main/java/io/avaje/inject/BeanScope.java
+++ b/inject/src/main/java/io/avaje/inject/BeanScope.java
@@ -5,6 +5,7 @@ import java.lang.reflect.Type;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.lang.Comparable;
 
 import org.jspecify.annotations.Nullable;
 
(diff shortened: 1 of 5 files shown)


mintdaniel42:
Sorry didn't think of toList() being a 16+ method, fixing it using collect()...

SentryMan:
Thanks for taking the time to contribute.

> now we are able to sort it when our interface implements Comparable without using any reflection or Runtime annotations

Since whatever interface you're looking for extends Comparable, wouldn't something like this be sufficient instead of adding another method?

```java
var list = scope.list(SomethingThatExtendsComparable.class);
list.sort(null);
```

mintdaniel42:
I added it because of https://github.com/avaje/avaje-inject/issues/723 but yes you are right 😄 sorry for overcomplicating things

rbygrave:
> I added it because of #723

But 723  only says ... **_"It would be cool if ..."_** - it does not actually give a use case of why it is desired or necessary.  As in, we shouldn't really add a feature because it's cool but more because we understand the use case that needs it (and that use case isn't best satisfied some other way).

Also note that this PR doesn't yet have a test that exercises the method and confirms it's behaviour.

PR #722, opened 2024-10-31
Add annotation for module-path `InjectPlugin` auto-detection

### LINKED ISSUE #724: Module-Path Detection doesn't work for test compile
As the title says, it seems during test compilation `getAllModuleElements` doesn't work.

To resolve, we can try to write an `avaje-module-dependencies.csv` in the compile stage so it can be picked up during the test-compile step.

SentryMan:
- adds a `@PluginProvides` annotation that the processor can autodetect via `getAllModuleElements` (not sure I like the name though, feel free to change it if you got a better one)
- writes plugin/module text files for reuse during test compilation 

resolves #724

### DIFF
--- a/blackbox-aspect/src/main/java/module-info.java
+++ b/blackbox-aspect/src/main/java/module-info.java
@@ -1,3 +1,5 @@
+import org.example.external.aspect.spi.AspectPlugin;
+
 module blackbox.aspect {
 
   exports org.example.external.aspect;
@@ -6,7 +8,6 @@ module blackbox.aspect {
   requires io.avaje.inject;
   requires io.avaje.inject.aop;
 
-  //remove this and compilation fails
-  provides io.avaje.inject.spi.InjectExtension with org.example.external.aspect.sub.ExampleExternalAspectModule;
+  provides io.avaje.inject.spi.InjectExtension with AspectPlugin, org.example.external.aspect.sub.ExampleExternalAspectModule;
 
 }
--- a/blackbox-aspect/src/main/java/org/example/external/aspect/PluginProvidedClass.java
+++ b/blackbox-aspect/src/main/java/org/example/external/aspect/PluginProvidedClass.java
@@ -1,0 +1,3 @@
+package org.example.external.aspect;
+
+public class PluginProvidedClass {}
--- a/blackbox-aspect/src/main/java/org/example/external/aspect/spi/AspectPlugin.java
+++ b/blackbox-aspect/src/main/java/org/example/external/aspect/spi/AspectPlugin.java
@@ -1,0 +1,22 @@
+package org.example.external.aspect.spi;
+
+import org.example.external.aspect.PluginProvidedClass;
+
+import io.avaje.inject.BeanScopeBuilder;
+import io.avaje.inject.spi.InjectPlugin;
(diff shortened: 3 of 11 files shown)


SentryMan:
@rbygrave yeah I'm good with this: (unless you have a better name for the annotation)

rbygrave:
> adds a @PluginProvides annotation 

Why do you want to add this? Why do you think we need this?

SentryMan:
> Why do you want to add this? Why do you think we need this?

With this, I can create internal plugin libraries that will be auto-detected by avaje inject without needing a maven/gradle plugin. I've got a couple of internal ones that this would be convenient to have. It is the natural extension of https://github.com/avaje/avaje-inject/issues/716.

SentryMan:
Currently, we need to use the `inject-maven-plugin` if:

- the application is using the maven compiler plugin `annotationProcessorPaths`
- The application has a module-info

This PR makes it so that the `inject-maven-plugin` is only required if:

- the application doesn't have a module-info _**and**_ the application is using the maven compiler plugin `annotationProcessorPaths`

As I like to make use of JPMS, this PR is very convenient for me

rbygrave:
> I can create internal plugin libraries 

Like what? Why are you creating plugins rather than using "normal dependencies"?  That is, I'm not really expecting many people to create plugins at all.


> without needing a maven/gradle plugin

People just need to register a service loader right, so they don't need a maven/gradle plugin for that? I don't get it?


> natural extension of https://github.com/avaje/avaje-inject/issues/716

I don't see that no, in that I'm really expecting people to use "normal dependencies" and only very few expert users will create plugins. So no I don't see that point.

SentryMan:
> Like what? Why are you creating plugins rather than using "normal dependencies"? 

A plugin I have is for one of our encryption libraries. It only exposes a single class so a plugin that provides a single instance is a natural choice. I have a similar structure to the jsonb/validator plugins to configure it.

> That is, I'm not really expecting many people to create plugins at all.

Yeah but for the people who do, this would be pretty convenient for their consumers.

SentryMan:
> People just need to register a service loader right, so they don't need a maven/gradle plugin for that? I don't get it?

Ah wait I think I might see the disconnect. I'm talking about the consumer of custom `InjectPlugin`s when I say they need a maven plugin.

Suppose I create a plugin `EncryptPlugin` and register it with the `ServiceLoader`

```java
public class EncryptPlugin implements InjectPlugin {

  @Override

    return new Class<?>[] {Encryptor.class};
  }

  @Override
   Encryptor e = ...
    builder.beans(e);
  }
}
```

When someone pulls in the maven dependency for this library and tries to use it like this:
```java
@Singleton
public class WireOther {
  Encryptor e;

  public WireOther(Encryptor e) {
    this.e = e;
  }
}
```

Compilation will fail if any of these are true

- the application is using the maven compiler plugin `annotationProcessorPaths`
- The application has a module-info

the reason is that the inject generator cannot read what classes the `EncryptPlugin` provides without help

rob-bygrave (approved):

SentryMan:
@rbygrave perchance can you deploy an RC when you find the time?

rbygrave:
Hmmm. https://s01.oss.sonatype.org is currently returning status code 503 ... but yeah will do it shortly hopefully.

PR #727, opened 2024-11-02
Rename `isAddBeanFor` to `isBeanAbsent`

SentryMan:
I feel this rolls off the tongue a little better.

- deprecated the old methods for removal in 11
- change generation to use the new name


rbygrave:
Hmmm, `isBeanAbsent` is kind of ok but also kind of misleading [when parentOverride is true].

SentryMan:
how about `canRegister`?

SentryMan:
10.5 is looking pretty stacked, this can probably be the last feature to add

rbygrave:
I don't think `canRegister()` is an improvement. I'd say `isBeanAbsent` was better than `canRegister`. I'm just not sure about `isBeanAbsent` yet.

PR #733, opened 2024-11-14
Merge Detected Services into `META-INF`

### LINKED ISSUE #731: Add all detected modules/plugins to generated `META-INF/services/io.avaje.inject.spi.InjectExtension` entries
So I was thinking about how I did https://github.com/avaje/avaje-inject/pull/704, and then it struck me. What if we do the same for all detected modules/plugins?

In this way, users would no longer need to manually configure gradle or the maven shade/assembly plugin to merge service files when using multi-module projects.

SentryMan:
- Now will add all detected modules/plugins to `META-INF/services`
- add an option to disable just in case
- now maven/gradle plugin will output plugin information in a csv format

resolves #731 

relies on prisms 1.35

### DIFF
--- a/inject-generator/pom.xml
+++ b/inject-generator/pom.xml
@@ -13,7 +13,7 @@
   <name>avaje inject generator</name>
   <description>annotation processor generating source code for avaje-inject dependency injection</description>
   <properties>
-    <avaje.prisms.version>1.34</avaje.prisms.version>
+    <avaje.prisms.version>1.35</avaje.prisms.version>
     <!-- VALHALLA-START ___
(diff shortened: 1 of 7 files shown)


rbygrave:
> add all detected modules/plugins to META-INF/services

I'm looking to understand the exact use case for this. I believe this is only applicable when the maven shade plugin is used (or similar) to merge multiple modules into a single jar right?

As I see it, there will likely be other META-INF/services that also need to be handled/merged by the maven shade plugin, so this doesn't address that issue.

So what is the exact use case for this?

rbygrave:
Ah, apologies, conversation is on the issue ... moving over there.

rbygrave (approved):

Issue #735, opened 2024-11-16
Investigate Background Bean Initialization


SentryMan:
It seems spring has added [background initialization](https://docs.spring.io/spring-framework/reference/core/beans/java/composing-configuration-classes.html#beans-java-startup-background) as a core DI feature.

Something like this would synergize well with beans that take some time to construct, allowing for other beans to be wired during the same time.

SentryMan:
My initial thought is to generate background beans like this

```java
@Generated("io.avaje.inject.generator")
public final class ExampleClass$DI  {

  public static void build(Builder builder) {
    if (builder.isAddBeanFor(Example.class)) {
      var future =
          CompletableFuture.supplyAsync(
              () -> new ControllerClass(builder.get(Dependency1.class, "!dependency1")));
      builder.registerProvider(future::join);
    }
  }
}
```

SentryMan:
Closing until we get actual real life usecases for such a feature

PR #736, opened 2024-11-17
Add Async Bean Generation

SentryMan:
First crack at #735 

given a class
```java
@Singleton
@AsyncBean
public class BackgroundBean {
  @Inject Provider<Integer> intProvider;
}
```

the following will generate

```java
public final class BackgroundBean$DI  {

    if (builder.isAddBeanFor(BackgroundBean.class)) {
      builder.registerProvider(CompletableFuture.supplyAsync(() -> {
        var bean = new BackgroundBean();
        bean.intProvider = builder.getProvider(Integer.class);
        return bean;
      })::join);  //use join to turn this future into a Provider 
    }
  }
}
```

### DIFF
--- a/inject/src/main/java/io/avaje/inject/spi/DBeanMap.java
+++ b/inject/src/main/java/io/avaje/inject/spi/DBeanMap.java
@@ -6,12 +6,8 @@ import jakarta.inject.Provider;
 
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.HashSet;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 
 /**
  * Map of types (class types, interfaces and annotations) to a DContextEntry where the
@@ -19,8 +15,8 @@ import java.util.Set;
  */
 final class DBeanMap {
   private static final Optional<Object> EMPTY = Optional.empty();
-  private final Map<String, DContextEntry> beans = new LinkedHashMap<>();
-  private final Set<String> qualifiers = new HashSet<>();
+  private final Map<String, DContextEntry> beans = new ConcurrentHashMap<>();
+  private final Set<String> qualifiers = Collections.synchronizedSet(new HashSet<>());
 
   private NextBean nextBean;
   private Class<? extends AvajeModule> currentModule;
--- a/inject/src/main/java/io/avaje/inject/spi/DBeanMap.java
+++ b/inject/src/main/java/io/avaje/inject/spi/DBeanMap.java
@@ -6,8 +6,12 @@ import jakarta.inject.Provider;
 
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
-import java.util.*;
-import java.util.concurrent.ConcurrentHashMap;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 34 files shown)


SentryMan:
what should be the annotation's name?

- `@Async`
- `@AsyncBean`
- `@Background`

SentryMan:
hmm, perhaps it should be an option of the `@Lazy` annotation rather than it's own annotation

inject/src/main/java/io/avaje/inject/spi/DBeanMap.java
```diff
@@ -6,21 +6,17 @@
...
 /**
  * Map of types (class types, interfaces and annotations) to a DContextEntry where the
  * entry holds a list of bean instances for that type.
  */
 final class DBeanMap {
   private static final Optional<Object> EMPTY = Optional.empty();
-  private final Map<String, DContextEntry> beans = new LinkedHashMap<>();
-  private final Set<String> qualifiers = new HashSet<>();
+  private final Map<String, DContextEntry> beans = new ConcurrentHashMap<>();
+  private final Set<String> qualifiers = Collections.synchronizedSet(new HashSet<>());
```
SentryMan, line 19: Do we require synchronization? The async beans don't have an opportunity to write to these collections. They are registered as jakarta providers.

rbygrave:
> Do we require synchronization?

No, I'm pretty sure I got that wrong (and we need to undo my changes
there). The wiring all happens single threaded in the foreground.

> Lazy

Id say thats "on demand". Is there benefit in "on demand + via another
thread"? ... cause the thread that invoked the lazy is just waiting? Hmm.

rbygrave:
> use case

The case I can think of is where an app has multiple db connection pools,
and there is a desire for these to initialize in parallel in the
background.

So the effective benefit is that initialisation is done in parallel (vs
currently serial).

... but to me this is not a case for Lazy per se [because I want it to fail
eagerly if any of those pools fail to initialise].

I should check the spring issue related to this.

SentryMan:
> I should check the spring issue related to this.

Apparently, it's one of the most voted issues ever: https://github.com/spring-projects/spring-framework/issues/13410

rbygrave:
> Apparently, it's one of the most voted issues ever

We have to be a little bit careful though because:
- some people are nuts
- some of those cases are ... "self inflicted" / avoidable 
- some of this is down to the Spring DI specific mechanism and not actually applicable to avaje-inject.

> Parallelize Component Scanning to Improve Spring Startup Time

avaje-inject avoids scanning altogether PLUS has already ordered the dependency wiring so yes we are a lot faster than Spring DI due to that. For myself, I also use libraries that also can avoid classpath scanning (Ebean ORM, ebean-migrations ... so no scanning and no dynamic proxies).

In the past I have been here though in terms of a Spring app starting up in around 30+ seconds. In a large modular application (not microservices, more "modular monolith") people can "feel" the startup time with Spring but again I'm saying we don't actually have this at all with avaje-inject because we don't have any scanning or any dynamic proxies.

So I'd say that slow startup time in an avaje-inject app is going to come from specific network activities (like obtaining remote configuration secrets/properties, initialising datasources, running database migrations, initialising queuing clients,  populating caches on startup).


> we have a startup time of out complete production environment of ~2h 30 Minutes.

That's just nuts and also provides zero detail as to what is actually slow. 


> Background initialization option for JPA EntityManagerFactory / Hibernate SessionFactory

That's due to Hibernate's use of classpath scanning and dynamic proxies. I don't have a lot of sympathy for people who do this to be honest.

SentryMan:
Seems like a solution in search of a problem. Closing unless somebody _actually_ needs it.

PR #739, opened 2024-11-20
Fix Repeatable Annotation Name Collision

SentryMan:
When using JDK 23+ module imports, the `List` type from `java.base`  and `io.avaje.inject.aop` have a name collision.

```java
import module java.base;
import module io.avaje.inject;

public class Example {

  void main() {
    List.of(); // this fails to resolve the import
  }
}
```


rbygrave (approved):
Nice catch !!

Issue #749, opened 2025-01-02
Feature request: Allow factory beans to have a @PreDestroy method to match/mirror a @Bean method


ascopes:
**Edit:** This is an edit to the summary as a TLDR on what this enhancement request has become.

The problem is specific to `@Factory` components that have methods that create `@Bean` components, and wanting to have a `@PreDestroy` method invoked on those beans.

The use case is that using Vertx the desired PreDestroy is actually a chained method invocation of `vertx.close().blockingAwait();`.

#### Change 1 - Change `@Bean(destroyMethod)` to support method chaining

So for example  `@Bean(destroyMethod="close().blockingAwait()")` now works. This change was added via https://github.com/avaje/avaje-inject/pull/754

#### Change 2 - Allow Factories to have a `@PreDestroy` method that matches a `@Bean` method

Factories naturally have `@Bean` methods to **_create_** components and it seems natural to allow those factories to have a `@PreDestroy` method that would match, for example:

```java
@Factory
final class MyFactory {

  @Bean 
  Vertx vertx() {
    var opts = new VertexOptions().setUseDaemonThread(true);
    return Vertx.vertx(opts);
  }
  
  @PreDestroy
  void destroy(Vertx vertx) {
    vertx.close().blockingAwait();
  }
}
```

This is supported via https://github.com/avaje/avaje-inject/pull/756

Also changing the title to reflect that this issue relates specifically to factories only. Normal `@Singleton` components can and should have their own `@PreDestroy` methods and shutdown logic (and should not have an extra dependencies required to execute the shutdown logic - aka normal `@PreDestroy` methods are not allowed to have arguments). 

-----------------

First off, thanks for the awesome annotation processor library! This has slotted really nicely into some work I am doing to make some components much easier to manage without the overhead and hassle of a full runtime CDI framework like Spring!

I have a small issue and I am wondering if a feature could be introduced to help address it.

## Background

In my application, I am making use of Vert.x RXJava3 components. I have some configuration resembling the following:

```java
import io.avaje.inject.Bean;
import io.avaje.inject.Factory;
import io.vertx.core.VertxOptions;
import io.vertx.rxjava3.core.Vertx;

@Factory
public class MyContext {
  @Bean 
  public Vertx vertx() {
    var opts = new VertexOptions().setUseDaemonThread(true);
    return Vertx.vertx(opts);
  }
}
```

[...]

PR #750, opened 2025-01-04
BeanScope PreDestroy

SentryMan:
Now can do 
```java
  @PreDestroy(priority = 100)
  void preDestroy(BeanScope scope) {
  
   // use the beanscope to close tricky beans

  }
```

- Adds a new interface for pre-destroy hooks that accept the current BeanScope
- refactor internals to convert `AutoCloseable` Interfaces for compatibility
- refactor postConstruct internals to convert `Runnable` to `Consumer<BeanScope>` instead of having two separate lists

Solves #749

### DIFF
--- a/blackbox-test-inject/src/main/java/org/example/myapp/HelloService.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/HelloService.java
@@ -92,7 +92,7 @@ public class HelloService {
   }
 
   @PreDestroy(priority = 100)
-  void preDest() {
+  void preDest(BeanScope scope) {
     MyDestroyOrder.add("HelloService");
   }
 }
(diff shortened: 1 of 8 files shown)


rbygrave:
I don't like this idea of `void preDestroy(BeanScope scope)` as we are invoking methods on a BeanScope that is in the process of "shutting down" - I feel there will be a lot of problematic cases / race conditions etc.

My preference would be to NOT support this. Currently there is no actual known use cases for PreDestroy methods that need to take the BeanScope right?

ascopes:
@rbygrave another option could be to allow the BeanScopeBuilder to be passed as a parameter to Bean methods. That'd allow registration via the destroy hook method.

ascopes:
Looking at https://docs.oracle.com/javase/8/docs/api/javax/annotation/PreDestroy.html it possibly violates these constraints too.

rbygrave:
> another option could be to allow the BeanScopeBuilder

I'm suggesting we try to stick to "First Principals" and first understand clearly the use cases that are not supported nicely/elegantly and the options for handling these cases. Currently, the ONLY use case we have at this stage is the Vertx one with a PreDestroy method that has method chaining `.close().blockingAwait()` [and we can actually support this by extending the existing mechanism - I have a PR ready to go for that].

We have no other use cases presented to us [yet]. That is, we currently have no use cases of a PreDestroy method that actually needs BeanScope. We currently have no use cases of a PreDestroy method that needs any parameters [and yes, as you point out if we did that would violate the standard there].

PR #751, opened 2025-01-04
Now can import `Prototype`/`Lazy` Beans

SentryMan:
Adds an attribute to the `Component.Import` annotation to make imported types have the prototype/lazy scope

### DIFF
--- a/inject/src/main/java/io/avaje/inject/Component.java
+++ b/inject/src/main/java/io/avaje/inject/Component.java
@@ -67,9 +67,10 @@ public @interface Component {
   @Target({TYPE, PACKAGE, MODULE})
   @interface Import {
 
-    /**
-     * Types to generate DI classes for.
-     */
+    /** Types to generate DI classes for. */
     Class<?>[] value();
+
+    /** Whether the imported types should have the prototype scope. */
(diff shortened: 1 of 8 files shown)


inject/src/main/java/io/avaje/inject/Component.java
```diff
@@ -67,9 +67,10 @@
...
   @interface Import {
 
-    /**
-     * Types to generate DI classes for.
-     */
+    /** Types to generate DI classes for. */
     Class<?>[] value();
+
+    /** Whether the imported types should have the prototype scope. */
+    boolean prototype() default false;
```
rbygrave, line 74: This would apply to all classes in this import.  That leads me to think that we likely need an additional ImportList annotation and a `@Repeatable(ImportList.class)` so that we can do something like:

```java
@Component.Import(value = SomeComponent.class)
@Component.Import(value = MyProtoComponent2.class, prototype=true)
```

rbygrave: Or alternatively a different annotation like `ImportPrototype`

rbygrave (approved):

PR #752, opened 2025-01-05
`PostConstruct` method Direct Injection

SentryMan:
generates the lookup and fetches beans for `PostConstruct ` lifecycle methods. (Before we could only inject the `BeanScope`)

Given:

```java
@Singleton
public class Minos {

  @PostConstruct
  void prepareThyself(Serpent serpent, Consumer<String> c, BeanScope b) {}
}
```

the following is generated:

```java
public final class Minos$DI  {

  public static final Type TYPE_ConsumerString =
      new GenericType<Consumer<String>>(){}.type();

    if (builder.isBeanAbsent(Minos.class)) {
      var bean = new Minos();
      builder.addPostConstruct(b -> $bean.prepareThyself(b.get(Serpent.class,"!serpent"), b.get(TYPE_ConsumerString), b));
    }
  }
}

### DIFF
--- a/blackbox-test-inject/src/main/java/org/example/myapp/config/LifeFour.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/config/LifeFour.java
@@ -1,0 +1,18 @@
+package org.example.myapp.config;
+
+import io.avaje.inject.PostConstruct;
+import jakarta.inject.Named;
+import jakarta.inject.Singleton;
+
+@Singleton
+public class LifeFour {
+
+  public String _state;
+
+  @PostConstruct
+  void post(@Named("foo") LifeOne one, LifeTwo two) {
+    _state = "post|"
+      + (one != null ? "one|" : "")
+      + (two != null ? "two" : "");
+  }
+}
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by rbygrave** [comment]
Why?  What is the use case where this makes sense to do?

**Comment by SentryMan** [comment]
Well, we already accept a bean scope as a parameter, this just eases the process of using the existing feature.

**Comment by SentryMan** [comment]
But answer you directly, a use case could be to send an event to any `@Observes' listeners.

**Comment by rbygrave** [comment]
> Well, we already accept a bean scope as a parameter, this just eases the process of using the existing feature.

Not quite in that the Builder isn't the same thing as the BeanScope. The BeanScope is the immutable collection of beans in the scope, and the Builder is expected to be lived for the duration of building the scope. That is, the side effect of this change is that there are lifecycle methods that register that keep a reference to the Builder.

**Comment by SentryMan** [comment]
> That is, the side effect of this change is that there are lifecycle methods that register that keep a reference to the Builder.

No? 
```java

 builder.addPostConstruct(b -> $bean.prepareThyself(b.get(Serpent.class,"!serpent"), b.get(TYPE_ConsumerString), b));
```

`b` here is the fully constructed beanscope, there's no reference to the builder

rbygrave:
> b here is the fully constructed beanscope, there's no reference to the builder.

Ah cool !!

rbygrave (approved):

PR #753, opened 2025-01-05
[Internals] Make generation easier to debug

SentryMan:
Modifies the Append Class to have a `toString` implementation using stringbuilder. In this way, we can easily debug file formatting/generation issues in a debugger

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ build/
 inject-generator/avaje-inject
 inject-generator/avaje-inject-generator
 inject-generator/avaje-processors.txt
+inject-generator/avaje-module-dependencies.csv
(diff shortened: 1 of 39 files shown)


rbygrave:
For debugging I tend to generate comments into the source code.  I'm kinda thinking that this potentially will add a decent amount of extra memory consumption/churn for everyone running the annotation processor [not just us]? Hmmm.

SentryMan:
> For debugging I tend to generate comments into the source code.

I like to debug the annotation processor with my IDE while it's running using the `InjectProcessorTest` test.

>  will add a decent amount of extra memory consumption

I'll make it a flag that's turned off by default then

rbygrave (approved):

PR #754, opened 2025-01-06
#749 Add support for chained methods in @Bean(destroyMethod)

rbygrave:
For example: `@Bean(destroyMethod = "reaper().stop()")`


SentryMan (approved):
LGTM

PR #756, opened 2025-01-06
#749 Add support for a @PreDestroy method on a factory, alternative to @Bean(destroyMethod)

rbygrave:
Example:
```java
@Factory
final class MyFactory {

  @Bean
  MyComponent myComponent() {
   ...
  }

  @PreDestroy(priority = 3000)
  void destroyMyComponent(MyComponent bean) {
    ...
  }
```

- ONLY valid on factory components
- An alternative to @Bean(destroyMethod)
- The @PreDestroy method matches to a @Bean by type only (ignores qualifiers, only 1 argument to the destroy method)

### DIFF
--- a/blackbox-test-inject/src/main/java/org/example/myapp/MyNestedDestroy.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/MyNestedDestroy.java
@@ -4,16 +4,16 @@ import java.util.concurrent.atomic.AtomicInteger;
 
 public class MyNestedDestroy {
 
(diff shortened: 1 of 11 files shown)


SentryMan:
While technically I see nothing wrong, I'm not a fan of being able to have multiple `@PreDestroy`  methods in a single class.

rbygrave:
> I'm not a fan of ...

Yes, it isn't perfect per se. I'm comfortable enough that it is intuitive enough and that we've limited this feature enough that it won't become a future regret. Time will tell.

PR #759, opened 2025-01-10
Remove implied qualifiers from PostConstruct Generation

SentryMan:
There's no need to generate implied qualifiers.

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
@@ -382,7 +382,7 @@ final class BeanReader {
       if (Constants.BEANSCOPE.equals(param.getFullUType().fullWithoutAnnotations())) {
         writer.append(beanScopeString);
       } else {
-        param.builderGetDependency(writer, builderName);
+        param.builderGetDependency(writer, builderName, !prototype);
       }
     }
     writer.append(")");
--- a/inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
@@ -548,6 +548,10 @@ final class MethodReader {
     }
 
     void builderGetDependency(Append writer, String builderName) {
+      builderGetDependency(writer, builderName, false);
+    }
+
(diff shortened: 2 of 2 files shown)


rbygrave:
> There's no need to generate implied qualifiers.

Why do you think that they are not required?

SentryMan:
Because if the bean has no qualifier,  `BeanScope.get` is sufficient.

rbygrave:
> Because if the bean has no qualifier

To me this is a vague statement because the qualifiers can be on **_both_** the supplier/bean ... and the location where that bean is injected.

When injecting a bean we don't know if the supplier(s) are qualified or not, hence the use of implied name.

When you say "Because if the bean has no qualifier" what do you mean? Which part are you talking about? How does the injecting part know if there are multiple suppliers with qualifiers or not?

SentryMan:
It's a fully constructed Beanscope, you don't need to do:

`BeanScope.builder().build().get(Something.class, "!something")` to retrieve a `Something` instant

SentryMan:
It's a fully constructed Beanscope, you don't need to do:

`BeanScope.builder().build().get(Something.class, "!something")` to retrieve a `Something` instant

rbygrave:
Because in the BeanScope there can be **_more than 1 instance_** for Something.class. So that just fails at runtime if that is the case.

SentryMan:
Suppose I have

```java
@Singleton
public class Minos implements Boss {

}
...

@Singleton
public class MinosPrime implements Boss {

}
```

what is the difference between 

```java
BeanScope.builder().build().get(Boss.class, "!something");
```
and
```java
BeanScope.builder().build().get(Boss.class);
```

would they both not fail with the same error?

SentryMan:
eh nvm

rbygrave:
Yeah, being explicit and changing your example above to:

```java
@Singleton // implied name of "Minos" 
public class MinosBoss implements Boss {

}
...

@Singleton // implied name of "MinosPrime"
public class MinosPrimeBoss implements Boss {

}
```

```java
beanScope.get(Boss.class, "!minos");  // works
beanScope.get(Boss.class);                   // error
```

SentryMan:
You learn something new every day

PR #760, opened 2025-01-13
Failing test for ModuleOrdering that skips the extra test module(s)

rbygrave:
Test components in src/test that are not annotated with @TestScope go into an extra module. This module is ignored/skipped by the compiled module ordering.

A fix for this is desired.

### DIFF
--- a/blackbox-aspect/pom.xml
+++ b/blackbox-aspect/pom.xml
@@ -5,7 +5,7 @@
   <parent>
     <artifactId>avaje-inject-parent</artifactId>
     <groupId>io.avaje</groupId>
-    <version>11.1-RC2</version>
+    <version>11.1</version>
   </parent>
   <modelVersion>4.0.0</modelVersion>
 
--- a/blackbox-other/pom.xml
+++ b/blackbox-other/pom.xml
@@ -6,7 +6,7 @@
   <parent>
     <artifactId>avaje-inject-parent</artifactId>
     <groupId>io.avaje</groupId>
-    <version>11.1-RC2</version>
+    <version>11.1</version>
   </parent>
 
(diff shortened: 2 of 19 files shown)


rbygrave:
The complied ModuleOrdering contains:

```java
  @Override
  public void add(AvajeModule module) {
    final var index = INDEXES.get(module.getClass().getTypeName());
    if (index != null) {
      sortedModules[index] = module;
    }
  }
```
... so modules that are not known are effectively lost.

rbygrave:
The fix here adds a supportsExpected method to ModuleOrdering like:

```java
public interface ModuleOrdering extends InjectExtension {

   /**
    * Return true if ordering supports the modules passed in.
    */
   default boolean supportsExpected(List<AvajeModule> modules) {
     return true;
   }
...
```
.. and changes the generated code to implement this with a check that the modules passed are the ones it expects to order.

SentryMan (approved):
LGTM, but the Compiled Ordering generation is an opt-in feature that's disabled by default, so I doubt this solves the actual issue that prompted this PR

SentryMan:
Wait, actually this might work

rbygrave:
> Compiled Ordering generation is an opt-in feature that's disabled by default, so I doubt this solves the actual issue that prompted this PR

Yes indeed - maybe there is another issue. I headed this way because it should have worked and I was able to reproduce the error this way. Hmm.

I'll include a test that wires a BeanScope in postConstruct(). Let me know if you are still happy with this perhaps.

SentryMan (approved):
Tests also pass when I disable the strict wiring so I'm fine with this, let's publish an RC and see if we get some feedback

PR #761, opened 2025-01-13
[test] allow list injection for tests

SentryMan:
now can do
```java
@InjectTest
class ListInject {

  @Inject List<Somei> testSomei;

  @Test
  void test() {

    assertThat(testSomei).isNotEmpty();
  }
}
```

### DIFF
--- a/inject-test/src/main/java/io/avaje/inject/test/MetaReader.java
+++ b/inject-test/src/main/java/io/avaje/inject/test/MetaReader.java
@@ -362,8 +362,13 @@ final class MetaReader {
         final var rawType = parameterizedType.getRawType();
         final var typeArguments = parameterizedType.getActualTypeArguments();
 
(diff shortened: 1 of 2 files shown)


inject-test/src/main/java/io/avaje/inject/test/MetaReader.java
```diff
@@ -349,8 +352,29 @@ Object get(Object instance) {
...
+          return;
+        }
+
+        if (rawType.equals(Optional.class)) {
+          set(field, beanScope.getOptional(typeArguments[0], name), testInstance);
+          return;
+        }
+      }
+
       if (!valueAlreadyProvided) {
```
rbygrave, line 376: Can this `if` get removed now?  ... cause it's checked with early return right at the top now

rbygrave (approved):

PR #775, opened 2025-03-11
[inject-test] Generate Method Handle Lookups

SentryMan:
Since mockito internals are now forbidden, generate a lookup class and use `VarHandle`s

- use `VarHandle` for getting/setting test fields
- adds a new processor to generate test lookups
- adds new SPI to inject test


rbygrave (approved):

rbygrave:
Nice work by the way!!

PR #778, opened 2025-03-16
Fix ExternalProvider on JDT

SentryMan:
turns out getDirectives throws an NPE if it's an automatic module. Only happens with JDT though


SentryMan:
After this. I think we're good to release 11.3

rbygrave (approved):

PR #781, opened 2025-03-25
Detect clash with multiple @Bean methods returning same type

rbygrave:
Given a `@Factory` with 2 or more bean methods that return the same type, and effectively clash on qualifier name (typically missing) then make this a compilation error.

The error suggests to add a `@Named` or qualifier annotation to resolve the issue.

Example:
```java
@Factory
class MyFactory {

  @Bean
  BFace one() {
    ...
  }

  @Bean
  BFace two() {
    ...
  }
```

### DIFF
--- a/blackbox-test-inject/src/main/java/org/example/myapp/config/BFace.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/config/BFace.java
@@ -1,0 +1,6 @@
+package org.example.myapp.config;
+
+public interface BFace {
+
+  String hi();
+}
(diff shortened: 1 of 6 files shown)


rbygrave:
An variation to this would be to give the `@Bean` a default qualifier based on the method name.

So for example:
```java
@Bean BFace one() { ...
```
... would get the default qualifier name of `one` from the method name.

SentryMan (approved):
is this a thing that happens?

rbygrave:
> is this a thing that happens?

I don't know if other people hit this, but I did hit it yesterday accidentally myself when I was looking to inject multiple Filter's via a Factory. I'm hoping people haven't hit it because it's a bit subtle and confusing when it occurs.

PR #789, opened 2025-03-31
Fix Jex Request-Scoped Context

SentryMan:
As the title says


rbygrave (approved):

SentryMan:
@rbygrave you'll have to change the branch protection rule

rbygrave:
> you'll have to change the branch protection rule

Meaning that you can't do something and you need it to change.

What can't you do?
How do you need it to change?

SentryMan:
So that JDK 11 is not required

PR #792, opened 2025-04-02
[test] remove hard dependency on events

SentryMan:
Saw strange behavior when I tried excluding the events dependency and running tests.


SentryMan:
hmm

rbygrave (approved):

Issue #793, opened 2025-04-02
Cross-/multi-scope bean wiring


jbarlin:
Hi there! I'm experimenting with avaje-inject version 11.4 and exploring ways to work with `@Scopes` in order to make a mono-repo with multiple deployments, which share a lot of common code by virtue of sharing a common core and because it was previously a monolith.

Ideally, there'd be an easy way to:
 * Annotate each component with a small number of annotations, which defines which deployments it's applicable to
 * Easily start the application with a small "main" method

I've tried a few things, but have gotten stuck on these paths (code examples+more detail in the expandables):

<details>
<summary>Declare scopes requiring other scopes, automatically including them</summary>

Using the following structure:

```java
@Scope
@InjectModule(provides = {BeanInModA.class}, strictWiring = true)
public @interface ModAScope {}

@ModAScope
public class BeanInModA{}

@Scope
@InjectModule(provides = {BeanInModB.class}, strictWiring = true)
public @interface ModBScope {}

@ModBScope
public class BeanInModB{}

@Scope
@InjectModule(requires = {ModAScope.class, ModBScope.class}, strictWiring = true)
public @interface CrossCutScope {}

@CrossCutScope
public class BeanCross{
    public BeanCross(final BeanInModA beanInModA, final BeanInModB beanInModB) {

    }
}
```

The result means you cannot just refer to the `CrossCutModule` when starting an entire `BeanScope` - you have to declare parent scopes (which is documented), or start the `BeanScope` with all modules declared in order. You can't pass them to `BeanScope.builder().modules(...)` out of order (e.g. `ModAScope, CrossCutScope, ModBScope`) as you'll end up with exception at startup ("`java.lang.IllegalStateException`: Injecting null for `com.example.BeanInModB name:!beanInModB` ..."). 

This is fine if you have a small number of scopes with a clear parent/child hierarchy, but if you have a large number of scopes/modules, or there's not a clear relationship between them, it becomes quite unmanageable.

---

</details>
<details>
<summary>Meta-annotating defined Scopes</summary>

Using the following structure:

```java
@Scope
@Target(TYPE)
@Retention(SOURCE)
@InjectModule(strictWiring = true)
public @interface ModAScope {}

@Scope
@Target(TYPE)
@Retention(SOURCE)
@InjectModule(strictWiring = true)
public @interface ModBScope {}

@ModAScope
@ModBScope
@Target(TYPE)
@Retention(SOURCE)
public @interface MetaScope {}

@MetaScope
[...]

PR #798, opened 2025-04-11
(Purposefully failing) Add more tests for multi-scope

cbarlin:
Related to #793

This is an expansion of #797 to include both the currently passing case, and a few cases that currently fail. It also expands further the number of modules, and makes it a non-linear graph.


SentryMan:
it strikes me that this might just be a bit of a runtime problem, that inject doesn't sort manually provided modules

PR #800, opened 2025-04-11
Multi-Scope Wiring

### LINKED ISSUE #793: Cross-/multi-scope bean wiring

SentryMan:
@cbarlin would this suffice?

With this, you would have to add the `requires` property in your `@InjectModule` annotations to control the order.

- now will order manually provided `AvajeModule`s
- will generate an `allRequiredModules` static method if the module has no external project scope dependencies and no transitive constructor params

Resolves #793

### DIFF
--- a/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
+++ b/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
@@ -1,11 +1,8 @@
 package org.multi.scope;
 
-import org.multi.moda.ModAModule;
-import org.multi.modb.ModBModule;
-
 import io.avaje.inject.InjectModule;
 import jakarta.inject.Scope;
 
 @Scope
-@InjectModule(requires = {ModAModule.class, ModBModule.class})
+@InjectModule(requires = {ModAScope.class, ModBScope.class})
 public @interface CrossCutScope {}
--- a/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
+++ b/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
@@ -1,0 +1,10 @@
+package org.multi.scope;
+
+import io.avaje.inject.InjectModule;
+import jakarta.inject.Scope;
+import org.multi.modb.BeanInModB;
+
+@Scope
+@InjectModule(requires = {ModAScope.class, ModBScope.class}, strictWiring = true)
+public @interface CrossCutScope {
+}
--- a/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
+++ b/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
@@ -4,7 +4,5 @@ import io.avaje.inject.InjectModule;
 import jakarta.inject.Scope;
 
 @Scope
-@InjectModule(
-    requires = {ModAScope.class, ModBScope.class},
-    provides = CrossCutScope.class)
+@InjectModule(requires = {ModAScope.class, ModBScope.class})
 public @interface CrossCutScope {}
--- a/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
+++ b/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
@@ -2,7 +2,6 @@ package org.multi.scope;
 
 import io.avaje.inject.InjectModule;
 import jakarta.inject.Scope;
-import org.multi.modb.BeanInModB;
 
 @Scope
 @InjectModule(requires = {ModAScope.class, ModBScope.class}, strictWiring = true)
--- a/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
+++ b/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
@@ -4,6 +4,7 @@ import io.avaje.inject.InjectModule;
 import jakarta.inject.Scope;
 
 @Scope
-@InjectModule(requires = {ModAScope.class, ModBScope.class}, strictWiring = true)
-public @interface CrossCutScope {
-}
+@InjectModule(
+    requires = {ModAScope.class, ModBScope.class},
+    provides = CrossCutScope.class)
+public @interface CrossCutScope {}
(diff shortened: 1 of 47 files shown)


cbarlin:
Looks great! Checking it out locally I can make it work with `strictWiring = true` on them as well (and make it work with a mix of some with and some without), if we wanted to add those to the test cases.

But yeah that solves the ordering problem nicely :)

cbarlin:
Thanks heaps @SentryMan! Did you want me to leave those other two PRs up?

blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
```diff
@@ -2,9 +2,9 @@
...
 import jakarta.inject.Scope;
-import org.multi.modb.BeanInModB;
 
 @Scope
-@InjectModule(requires = {ModAScope.class, ModBScope.class}, strictWiring = true)
-public @interface CrossCutScope {
-}
+@InjectModule(
+    requires = {ModAScope.class, ModBScope.class},
+    provides = CrossCutScope.class)
```
rbygrave, line 9: To me the `provides` should **_not_** be needed ultimately. To me we've needed to add them here ... but ultimately that wouldn't be desirable.

blackbox-multi-scope/src/test/java/org/multi/crosscut/BeanCrossTest.java
```diff
@@ -1,24 +1,97 @@
...
 
-import static org.assertj.core.api.Assertions.assertThat;
+import io.avaje.inject.BeanScope;
+import io.avaje.inject.spi.AvajeModule;
 
 class BeanCrossTest {
 
-  @Test
-  void bootstrap() {
+  private static Stream<Arguments> allModulesDefined() {
```
rbygrave, line 32: IMO these tests are not what we ergonomically desire/need.

That is, we want to say "Give me all the modules that I need for CrossCutModule [and ideally give them in the correct order]" ... and use those to create a BeanScope.

These tests are more just ... we can order the given list of modules correctly.

SentryMan: > These tests are more just ... we can order the given list of modules correctly.

do we have an existing test for that?

blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
SentryMan: Would auto-providing the scope annotation suffice?

SentryMan:
Now you need only the requires property

cbarlin:
Looks awesome - thank you so much!

rbygrave (approved):

PR #802, opened 2025-04-12
Attempt multi scope no deps

cbarlin:
An attempt at building off of @SentryMan's `allRequiredModules` method...

It doesn't work for modules that have dependencies though, and I couldn't easily see a way to automatically bail if that was the case. It does work for the `blackbox-multi-scope` though... 

Sorry, wasn't sure where to put this attempt, and boiling down the changes I made to write in a ticket would probably be difficult.

### DIFF
--- a/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
+++ b/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
@@ -1,11 +1,8 @@
 package org.multi.scope;
 
-import org.multi.moda.ModAModule;
-import org.multi.modb.ModBModule;
-
 import io.avaje.inject.InjectModule;
 import jakarta.inject.Scope;
 
 @Scope
-@InjectModule(requires = {ModAModule.class, ModBModule.class})
+@InjectModule(requires = {ModAScope.class, ModBScope.class})
 public @interface CrossCutScope {}
--- a/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
+++ b/blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
@@ -4,7 +4,5 @@ import io.avaje.inject.InjectModule;
 import jakarta.inject.Scope;
 
 @Scope
-@InjectModule(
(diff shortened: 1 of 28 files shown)


inject-generator/src/main/java/io/avaje/inject/generator/SimpleModuleWriter.java
```diff
@@ -102,7 +106,7 @@ void write() throws IOException {
   private void writeRequiredModules() {
 
     ProcessingContext.modules();
-    Set<String> requiredModules = new TreeSet<>();
+    List<String> requiredModules = new ArrayList<>();
```
cbarlin, line 109: Probably not needed - I ran into an issue where it was complaining about missing required scopes, so was trying a few things until I made the change to have the scope auto-provide itself

blackbox-multi-scope/src/main/java/org/multi/scope/CrossCutScope.java
```diff
@@ -1,11 +1,8 @@
...
 
-import org.multi.moda.ModAModule;
-import org.multi.modb.ModBModule;
-
 import io.avaje.inject.InjectModule;
 import jakarta.inject.Scope;
 
 @Scope
-@InjectModule(requires = {ModAModule.class, ModBModule.class})
+@InjectModule(requires = {ModAScope.class, ModBScope.class})
```
cbarlin, line 7: These being modules or scopes is a hard one by the looks:
 * If they are modules, then we don't need to add to the autoProvides (although it works without this PR without doing that 🤔 )
 * If they are scopes, you can easily find the scopes they rely on

But then, either way, it (currently) falls down when you attempt to extract the module dependencies to build the method... So maybe having them as modules is better? Although then you may be trying to read something from the annotation that doesn't yet exist and I've ran into that before as a problem in another project...

SentryMan: so modules in the annotation doesn't fly because the module classes don't exist and the processor can't read them

SentryMan:
I've pulled your changes into #800

PR #805, opened 2025-04-15
Add more detections to BeanReader#shouldDelay

### LINKED ISSUE #804: Bean implementing generated interface isn't wireable
Hi there!

Using avaje 11.4, if we have the following setup:

```java
@NonAvajeAnnotation
class MyClassA {}

@Singleton
class MyClassB implements SomeGeneratedInterface {}

@Singleton
class MyClassC {
    public MyClassC (final SomeGeneratedInterface nonAvajeCreatedInterface) {

    }
}
```

In which the `SomeGeneratedInterface` is an interface generated by the `NonAvajeAnnotation`, then the app will fail to compile because `avaje-inject` cannot find a `SomeGeneratedInterface` to wire into `MyClassC`. This occurs even if `@BeanTypes({SomeGeneratedInterface.class})` is placed on `MyClassB`.

Looking at the annotation processing rounds, it appears that `avaje-inject` correctly delays `MyClassC` until the end of compilation because it can't find the dependencies, but the `MyClassB` isn't delayed and a `MyClassB$DI` is generated immediately without knowledge of `SomeGeneratedInterface`. Interestingly, I don't get an error about `MyClassB` not implementing `SomeGeneratedInterface`.

cbarlin:
Should resolve #804 by adding more detections to the `shouldDelay` method. Namely:
 * Check the `BeanTypes` annotation
 * Check for interfaces
 * Check for any super type


SentryMan (approved):
Changes look fine to me

inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
```diff
@@ -104,18 +105,38 @@ private boolean shouldDelay() {
...
+      .map(TypeMirror::getKind);
 
-    return Stream.concat(constructFields, methods)
-      .map(Element::asType)
-      .anyMatch(t -> t.getKind() == TypeKind.ERROR);
+    var constructorField = Stream.concat(construct, fields);
+    var methodInterfaces = Stream.concat(methods, interfaces);
+    var superclassBeanTypes = Stream.concat(superClass, beanTypes);
+    return Stream.concat(constructorField, Stream.concat(methodInterfaces, superclassBeanTypes))
+      .anyMatch(TypeKind.ERROR::equals);
```
cbarlin, line 139: Should this also check if `TypeKind` is one that has type arguments and that those are also not `TypeKind.ERROR`?

SentryMan: I feel like that's a niche enough issue that we don't have to slow down generation for it. @rbygrave what say you?

rbygrave: > niche enough issue that we don't have to slow down generation for it

Yes, I agree with that.  We can go with this, and then if/when that need arises we can revisit this. This is also along the lines _"keep is as simple as possible for as long as possible"_. 

-----
Side note: I think we could look to use Stream.concat(... list ...) here. I might see what that looks like.

rbygrave: Ok, can I get you guys to check those last 2 commits?  I think @SentryMan has already checked that main one there where everything returns as `Stream<TypeKind>` ... and the last one was just style really?

So we good to merge?

SentryMan (approved):

inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
rbygrave: Nice simplification there. Ok, merging ...

rbygrave:
FYI: This is in version `11.5-RC3` which is on its way to central

cbarlin:
Awesomes - thank you both!

PR #806, opened 2025-04-16
For Java 21+ we get WARN saying mockito agent should be explicit command line arg

rbygrave:
So hmm, should this go into our parent pom for surefire?

Refer to maven section in:
https://javadoc.io/doc/org.mockito/mockito-core/latest/org.mockito/org/mockito/Mockito.html#0.3


SentryMan:
I say we put `-XX:+EnableDynamicAgentLoading` in the parent pom

rbygrave:
>  -XX:+EnableDynamicAgentLoading in the parent pom surefire when the JDK is > 21

Yeah, thats better. closing this.

PR #816, opened 2025-05-07
Support Array Injection

SentryMan:
Does _not_ collect beans into an array, but injects array instances. 

Also doesn't support generic arrays (I just didn't feel like it)

### DIFF
--- a/blackbox-test-inject/src/main/java/org/example/myapp/arrays/ArrayFactory.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/arrays/ArrayFactory.java
@@ -1,0 +1,20 @@
+package org.example.myapp.arrays;
+
+import io.avaje.inject.Bean;
+import io.avaje.inject.Factory;
+
+@Factory
+public class ArrayFactory {
+
+  @Bean
+  byte[] bytes() {
+    return new byte[] {};
+  }
+
+  @Bean
+  ArrayType[] type() {
+    return new ArrayType[] {};
+  }
+
+  public static class ArrayType {}
+}
--- a/blackbox-test-inject/src/main/java/org/example/myapp/arrays/Arrays.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/arrays/Arrays.java
@@ -1,0 +1,13 @@
+package org.example.myapp.arrays;
+
+import io.avaje.inject.Component;
+
+@Component
+public class Arrays {
+
+  byte[] arr;
+
+  public Arrays(byte[] arr) {
+    this.arr = arr;
+  }
+}
--- a/blackbox-test-inject/src/test/java/org/example/myapp/arrays/ArraysTest.java
+++ b/blackbox-test-inject/src/test/java/org/example/myapp/arrays/ArraysTest.java
@@ -1,0 +1,22 @@
+package org.example.myapp.arrays;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+import org.example.myapp.arrays.ArrayFactory.ArrayType;
+import org.junit.jupiter.api.Test;
+
+import io.avaje.inject.test.InjectTest;
+import jakarta.inject.Inject;
+
+@InjectTest
+class ArraysTest {
+  @Inject byte[] arr;
+  @Inject ArrayType[] arr2;
+
+  @Test
+  void test() {
+
+    assertNotNull(arr);
+    assertNotNull(arr2);
+  }
+}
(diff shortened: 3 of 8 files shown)


rbygrave:
Ok, I'll have another look at this later but I don't think this is the path / requirement.  I think instead this was about changing the Builder to have an array() method similar to the existing list() and set() methods and thus allowing a "collection" to be injected as an array (vs list or set).

That is, the type being registered isn't an actual array.

... but yeah I'll look again later and think some more.

SentryMan:
>  allowing a "collection" to be injected as an array (vs list or set).

I don't think even spring does this. What utility would something like that be when you can so easily convert a list into an array?

SentryMan:
>  changing the Builder to have an array() method

If you want to do something like that, an @QualifiedArray annotation or something can be added as a followup to behave in that manner.

rbygrave:
> I don't think even spring does this. 

Yeah it does. For example I see is Spring injecting an`Callback[]` for Flyway (which takes varargs).

> What utility would something like that be when you can so easily convert a list into an array?

Minimal utility value ... which is why I didn't bother. I don't think anyone has actually asked for it until just now but only then as a workaround. So yeah, it's a super rare case.

SentryMan:
I'm saying it's much more likely that someone would want to wire an array _instance_ than to collect beans with an array

rbygrave:
> I'm saying it's much more likely that someone would want to wire an array instance than to collect beans with an array.

I've seen 0 occurrences of "want to wire an array instance" 
I've seen 1 1/2 occurrences of "collect beans with an array" (the 1/2 given it was mentioned as a possible workaround) 

So 1.5 > 0.

... regardless "wire an array instance" isn't something I have any desire for avaje-inject to support so we wait for the actual person with the real use case to request that and then evaluate that request on it's merit (but we basically want to avoid array types in the underlying context map so ... its not a given that will change).

SentryMan:
> we basically want to avoid array types in the underlying context map

Is not an array an object like any other? Is there a specific reason to avoid array types?

rbygrave:
>  like any other [object]?

Its basically not the job of DI to inject ANY object ... and specifically not "Data". That is, the job of DI is really to inject "objects that represent Business Logic".

In this sense, we don't really expect to inject `String` or `Integer` or primitives or `byte[]` if we see this as "Data" ... but instead think of this "Data" as "External Configuration" [so avaje-inject explicitly doesn't have an equivalent to Spring `@Value`]

If we put a `Valhalla + "Record types" = data` perspective ... we don't actually expect DI to inject any type that is a Valhalla Value type or record type or primitive. If we find that we are ... we are more thinking that's probably "external configuration".


> Is not an array an object like any other? 

Sort of but also not really.  For example primitive arrays like `int[], byte[]` ... we'd see that as **_data_**.

If we have `MyType[]` ... it's more like a "collection of MyType" presented as an array.  If we want `MyType[]` to be a distinct type and NOT a kinda collection then we can easily wrap that in another type.


Does any of that make sense?

SentryMan:
> Its basically not the job of DI to inject ANY object ... and specifically not "Data".

Where does this "specifically not for data" requirement come from? I've never heard of this.

> That is, the job of DI is really to inject "objects that represent Business Logic".

This makes intuitive sense, and that's how it would be used in most cases. If you start with the premise that data should not be injected, the rest follows naturally. However, I'm not sold on the no data injection part.

> we don't really expect to inject String or Integer or primitives

We had a [request for primitives quite a while ago](https://github.com/avaje/avaje-inject/pull/300), so it seems strange to support wiring a `byte` but not a `byte[]`.

> Does any of that make sense?

The logic is valid, but I can't say I'm a fan of failing compilation when someone decides to use an array.

rbygrave:
> Where does this "specifically not for data" requirement come from? I've never heard of this.

I haven't heard it explicitly come from anyone else apart from me.

That is, I had been thinking about DI pretty deeply for a lot of years (like 20+ years of using DI). This particular insight comes from seeing other people try to explain DI very badly, plus comparing DI to Functional programming and how functions are composed there (direct parallels to DI), plus the design decisions in avaje-inject around `@Value`.

So not many people have done that amount of thinking on this topic. The direct parallels to FP function composition are maybe the most explicit due to the clear separation between functions and data.


> request for primitives ...

Yeah, we might regret that one day (e.g. valhalla). My take is that people merge external configuration and DI concepts, and request scoped injection also blurs it for people. Once Valhalla lands people will more explicitly separate "data" from "logic" in their mental models (e.g. value record types).

PR #817, opened 2025-05-08
Support Nested Wildcard Lists

SentryMan:
- fix cases where a list of generic wildcards generates bad code
- now generic types are registered with their main type as a key


rbygrave (approved):
Note I reverted the autoProvides() change

SentryMan:
> Note I reverted the autoProvides() change

Any particular reason why?

rbygrave:
> Any particular reason why?

For `MyType<T>` we don't actually know what type we can provide, unlike say `MyType<OtherType>`.  If another module needs `MyType<Foo>` we don't know if this module can provide that dependency or not.

For this reason these types have been excluded from the autoProvides.

PR #823, opened 2025-05-29
Fix TYPE_USE annotated dependencies

SentryMan:
Fix issue where TYPE_USE annotations would cause the wrong type to be registered, causing errors like:

`No dependency provided for java.util.concurrent.atomic.@org.jspecify.annotations.Nullable AtomicBoolean on org.example.myapp.lazy.LazyBean`

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/Util.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/Util.java
@@ -1,5 +1,13 @@
 package io.avaje.inject.generator;
 
+import static java.util.function.Predicate.not;
+import static java.util.stream.Collectors.toList;
+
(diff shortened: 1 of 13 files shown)


inject-generator/src/main/java/io/avaje/inject/generator/Util.java
```diff
@@ -223,7 +223,7 @@ static String extractMap(String rawType) {
   }
 
   static UtilType determineType(TypeMirror rawType, boolean beanMap) {
-    return UtilType.of(rawType.toString(), beanMap, rawType);
```
SentryMan, line 226: problem was that we weren't sanitizing the type mirror to string

rbygrave (approved):

PR #826, opened 2025-06-08
fix: handle nested java types like Flow.Publisher in Util.shortName

VincentPolfliet:
This PR fixes a bug in Util.shortName(...) where nested types like java.util.concurrent.Flow.Publisher were resolved incorrectly as Publisher, which caused generated code to fail compilation when the short name was not imported correctly.

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/Util.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/Util.java
@@ -141,26 +141,26 @@ final class Util {
     final int p = fullType.lastIndexOf('.');
     if (p == -1) {
       return fullType;
-    } else if (fullType.startsWith("java")) {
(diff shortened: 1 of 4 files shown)


SentryMan (approved):
Thanks for the contribution.

SentryMan:
> Nice, but there is **_insufficient test coverage_** here and this PR should NOT have been merged without that coverage @SentryMan

Is there something missing? Looks 100% to me
image

The problem described was not about wiring, but the generation of invalid code. We already have tests for wiring things with `GenericType`

rbygrave:
> Is there something missing? Looks 100% to me with all code paths hit.

Yeah I was wrong. Test coverage there was ok. I was concerned about missing coverage on the java types but was wrong there.

PR #829, opened 2025-06-09
Test only followup for #826, add test to blackbox-test-inject

rbygrave:
- Move the test objects GenericImpl, GenericInterfaceObject to blackbox-test-inject
- Long term it is not sufficient to compile but actually have a test that confirms it wires
- Also tidy UtilTest to merge the tests for shortName() + add some java types to that test


SentryMan (approved):
> Long term it is not sufficient to compile but actually have a test that confirms it wires

not totally sold on this for non-wiring related issues but ok

PR #851, opened 2025-07-02
Merge requires/provides with their auto variants

SentryMan:
Honestly, there is no reason to split into three separate methods, given that we support generic wiring.

- deprecates auto requires/provides
- generate only provides/requires
- move the generation order such that the build method is seen first. (got tired of scrolling past the massive classes/provides to see the wiring logic)


SentryMan:
> Their source is different.

Perhaps, but `AvajeModule#classes` has always been the recommended option to reflect on types, and it makes no such distinction.

rbygrave (approved):

PR #852, opened 2025-07-05
Determine Priority Annotations at Compile Time

### LINKED ISSUE #854: Revisit `@Priority`
The current implementation of `@Priority` and `BeanScope#listByPriority` presents several critical issues, leading to **confusing behavior** and **limiting its practical application**, especially in scenarios involving multiple modules and complex bean hierarchies.

---

The core problem lies in the **inconsistent and unexpected interactions** between `@Priority`, `@Primary`, `@Secondary`, and the `BeanScope` methods, particularly `listByPriority` and bean injection. This makes it difficult to reliably control bean selection and ordering.

**Specific Issues:**

1.  **Limited `@Priority` Placement:** The `@Priority` annotation **cannot be applied to factory beans**. This significantly reduces its utility for scenarios where bean creation is managed by factories.
2.  **`BeanScope#listByPriority` Ignores `@Primary` / `@Secondary`:** When using `BeanScope#listByPriority`, the order **does not respect** the explicit preference provided by `@Primary` or `@Secondary` annotations. This undermines the intent of these annotations for bean disambiguation.
3.  **`@Priority` Ineffectiveness in Wiring:** When multiple beans of the same type are available, `@Priority` has **no effect** on which bean is injected. This leads to unpredictable bean resolution during dependency injection.
4.  **Runtime Annotation Reading:** As a compile-time framework that uses annotation processing, reading `@Priority` annotations at **runtime** to extract their value runs counter to the idea of the entire project. This information should be processed during **compilation**.
5. I cannot stress enough how much the runtime annotation reading _personally annoys me_ now that it was pointed out
---

###  Desired Behavior

We expect `@Priority` to provide a **consistent and intuitive mechanism** for ordering beans and influencing bean selection during injection and when using `BeanScope#listByPriority`. Specifically:

* `@Priority` should be **applicable to factory beans**.
* `BeanScope#listByPriority` should **respect and integrate** with `@Primary` and `@Secondary` annotations to provide a consistent ordering.
* `@Priority` should **influence bean selection** during dependency injection when multiple beans of the same type are present, potentially in conjunction with `@Primary` and `@Secondary`.
* Annotation processing for `@Priority` should occur at **compile-time**.
---

Consider the following scenarios, which demonstrate the problematic behavior:

#### Scenario 1: `@Priority` and Wiring Inconsistency

Imagine a setup where different modules contribute `LobbyController` implementations, and we want to prioritize `SMPShardLobbyController`.

```java
@Slf4j
@Singleton
@Priority(1)
public final class SMPShardLobbyController extends LobbyController {
    @Inject
    public SMPShardLobbyController(
    ) {
        super();
    }

    @Override
    public boolean canTeleport(Player player) {
        log.info("SMPShard Can Teleport");
        return super.canTeleport(player);
    }
}
```

```java
@Singleton
@Priority(3)
public class LobbyController {
    @Inject
    public LobbyController(

    ) {

    }

    public boolean teleport(final Player player) {
      // impl
    }

    /**
     * Check whether a player can teleport to the lobby or not.
     * @param player the player
     * @return true if they can teleport false otherwise.
     */
    public boolean canTeleport(final Player player) {
        return true;
    }
}
```

```java
@Singleton
public final class LobbyCommand implements BaseCommand {
    private final LobbyController lobbyController;

    @Inject
    public LobbyCommand(
        final LobbyController lobbyController
    ) {
        this.lobbyController = lobbyController;
    }

    @Command("lobby|hub|l")
    @CommandDescription("Bring yourself back to the lobby")
    public void lobby(final PlayerSource source) {
        lobbyController.teleport(source.source());
    }
}
```

Problem: `BeanScope#get(LobbyController.class)` can often not return the `SMPShardLobbyController` as desired, even with `@Priority(1)`

#### Scenario 2: `BeanScope#listByPriority` doesn't respect `@Primary` or `@Secondary`

Consider a situation with `Gson` beans provided by different factories, using `@Primary` and `@Secondary` for disambiguation.

```java
@Slf4j
@Factory
public final class GsonProvider {
    private final Environment environment;

    @Inject
    public GsonProvider(
        final Environment environment
    ) {
        this.environment = environment;
    }

    @Primary @Bean
    @Named("base")
    public Gson provideGson() {
        return new GsonBuilder()
            /* impl */
            .create();
    }
}
```
```java
@Factory
public final class LayerGsonProvider {
    private final Gson baseGson;

    @Inject
    public LayerGsonProvider(
        @Named("base") final Gson gson
    ) {
        this.baseGson = gson;
    }

    @Secondary @Bean
    @Named("bukkit")
    public Gson layerGson() {
        return this.baseGson
            .newBuilder()
            .registerTypeAdapter(Location.class, new LocationSerializer())
            .create();
    }
}
```

Problem: When calling `BeanScope#listByPriority(Gson.class)`, the returned list does not reflect the order implied by `@Primary` or `@Secondary`. The base `Gson` (marked `@Primary`) and bukkit `Gson` (marked `@Secondary`) will not necessarily appear in the expected priority order.

SentryMan:
- read the `@Priority` annotations at compile time
- listByPriority will now sort `@Primary`/`@Secondary` beans
- simplify duplicate type matching logic
- Add support for using `@Priority` on factory `@Bean` methods

Resolves #854

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
@@ -37,6 +37,7 @@ final class MethodReader {
   private final boolean prototype;
   private final boolean primary;
   private final boolean secondary;
+  private final Integer priority;
   private final boolean lazy;
   private final boolean proxyLazy;
   private final TypeElement lazyProxyType;
@@ -68,6 +69,7 @@ final class MethodReader {
       prototype = PrototypePrism.isPresent(element);
       primary = PrimaryPrism.isPresent(element);
       secondary = SecondaryPrism.isPresent(element);
+      priority = Util.getPriority(element);
       lazy = LazyPrism.isPresent(element) || LazyPrism.isPresent(element.getEnclosingElement());
       conditions.readAll(element);
       this.lazyProxyType = lazy ? Util.lazyProxy(element) : null;
@@ -76,6 +78,7 @@ final class MethodReader {
       prototype = false;
       primary = false;
       secondary = false;
+      priority = null;
       lazy = false;
       this.proxyLazy = false;
       this.lazyProxyType = null;
@@ -273,6 +276,8 @@ final class MethodReader {
       writer.append(".asPrototype()");
     } else if (secondary) {
       writer.append(".asSecondary()");
+    } else if (priority != null) {
+      writer.append(".asPriority(%s)", priority);
     }
 
     if (proxyLazy) {
@@ -331,6 +336,8 @@ final class MethodReader {
         writer.append(".asPrimary()");
       } else if (secondary) {
         writer.append(".asSecondary()");
+      } else if (priority != null) {
+        writer.append(".asPriority(%s)", priority);
       } else if (prototype) {
         writer.append(".asPrototype()");
       }
@@ -526,7 +533,7 @@ final class MethodReader {
   }
 
   boolean isUseProviderForSecondary() {
-    return secondary && !optionalType && !Util.isProvider(returnTypeRaw);
+    return (secondary || priority != null) && !optionalType && !Util.isProvider(returnTypeRaw);
   }
 
   boolean isPublic() {
--- a/inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
@@ -4,6 +4,7 @@ import static io.avaje.inject.generator.APContext.logWarn;
 import static io.avaje.inject.generator.Constants.CONDITIONAL_DEPENDENCY;
 import static io.avaje.inject.generator.ProcessingContext.asElement;
 
+import java.lang.reflect.Type;
 import java.util.ArrayList;
 import java.util.Collections;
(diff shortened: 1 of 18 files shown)


inject-generator/src/main/java/io/avaje/inject/generator/MethodReader.java
```diff
@@ -526,7 +533,7 @@ boolean isLazy() {
   }
 
   boolean isUseProviderForSecondary() {
-    return secondary && !optionalType && !Util.isProvider(returnTypeRaw);
+    return (secondary || priority != null) && !optionalType && !Util.isProvider(returnTypeRaw);
```
rbygrave, line 536: So if it's got a `@Priority` also wire as `Provider<T>` which has the effect of only instantiating when needed.

Needed by .. being the highest priority or BeanScope#list or injecting a List/Set/Map of that type.

SentryMan: Indeed, the appearance of a priority annotation invites the possibility that a higher priority bean might exist

rbygrave (approved):
I really like this PR.

I think it's nice to merge the Priority in with Primary and Secondary.

PR #863, opened 2025-07-17
[test] better name for the refresh scope flag

SentryMan:


SentryMan:
after this let's create an RC

rbygrave (approved):

inject-test/src/main/java/io/avaje/inject/test/MetaInfo.java
```diff
@@ -54,7 +54,7 @@ private TestBeans buildTestBeans(GlobalTestBeans.Beans parent, Object testInstan
     String[] profiles = injectTest.map(InjectTest::profiles).orElse(new String[0]);
 
     if (profiles.length > 0
-        || injectTest.map(InjectTest::scopePerMethod).orElse(false)
+        || injectTest.map(InjectTest::refreshScope).orElse(false)
```
rbygrave, line 57: Should be scopePerTest ?

PR #878, opened 2025-07-30
DBuilder contains checking parent if present

cbarlin:
Apparently when I did #873 I missed that `DBuilder` didn't check the parent scope when calling `contains`... sorry!


rbygrave:
Are we missing test coverage on this?

rbygrave (approved):

cbarlin:
@SentryMan Thank you so much!

PR #889, opened 2025-08-22
Adds Compiler Warning to useless `@Inject`

SentryMan:
Now will warn when the `@inject` annotation is not needed


rbygrave:
I don't understand the motiivation.  Some people use `@Inject` on a constructor just to be more explicit and that would now be a warning?

What was the motivation for this?

SentryMan:
>  Some people use @Inject on a constructor just to be more explicit and that would now be a warning?

It does not affect anything, so to me it's a useless addition. Even so, if you don't like it that's that (I don't care about this _that_ much)

PR #896, opened 2025-10-01
Fix Proxy Generation with Final Methods

### LINKED ISSUE #895: Lazy fails to compile
Hi, it's great to hear there's compile time lazy but this breaks for some classes. Below is an example with SecureRandom. Same happens to e.g. `jooq` and some other libraries. It works fine with `11.5` but breaks with `11.6` and above.

`  /t0/app/build/generated/sources/annotationProcessor/java/main/org/example/SecureRandom$Lazy.java:16: error: type Provider does not take parameters
    private final Provider<SecureRandom> onceProvider;
`

The code is just

```java
@Lazy
@Factory
public class RandomFactory {
    @Bean
    public SecureRandom secureRandom() {
        return new SecureRandom();
    }
}
```

SentryMan:
Fixes issue where we would try to generate proxies for classes with final methods.

resolves #895

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,11 +6,19 @@ updates:
       interval: weekly
     open-pull-requests-limit: 10
     ignore:
-       - dependency-name: "jakarta.inject:jakarta.inject-api"
+      - dependency-name: "jakarta.inject:jakarta.inject-api"
+      # --- Configuration to ignore all JUnit dependencies (e.g., 5.x) ---
+      # Ignores all artifacts with Group ID 'org.junit.jupiter' (e.g., api, engine, params)
+      - dependency-name: "org.junit.jupiter:*"
+      # Ignores all artifacts with Group ID 'org.junit.platform' (e.g., launcher, suite)
+      - dependency-name: "org.junit.platform:*"
+      # Ignores the JUnit Vintage Engine
+      - dependency-name: "org.junit.vintage:*"
+      # -----------------------------------------------------------------------
     groups:
       dependencies:
         patterns:
-        - "*"
+          - "*"
     labels:
       - "dependencies"
     target-branch: "master"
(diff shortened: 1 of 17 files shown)


SentryMan:
Actually I don't see, because when I move it to blackbox it doesn't fail

rbygrave:
So when it's:
```java
  @Bean
  public Random secureRandom() {
    return new SecureRandom();
  }
```
Then it becomes a `registerLazy()` [beanReader.proxyLazy()] ... and we get a `Random$Lazy` generated and that doesn't support the protected method.
```java
 ...
      var factory = builder.get(RandomFactory.class);
      builder.registerLazy(() -> {
      return factory.secureRandom();
      }, Random$Lazy::new);    }
```
... but with SecureRandom its just a lazy [beanReader.lazy()] ... so no `Random$Lazy` generated 
```java
      var factory = builder.get(RandomFactory.class);
      builder.registerProvider(() -> {
      return factory.secureRandom();
          });
```

SentryMan:
well, it appears the bigger problem is situations where the no-arg constructor calls the methods. I'll just an escape hatch to disable proxy generation

SentryMan:
That's exactly right, it's the old behavior. In the case of random it doesn't appear that we can cleanly lazy it with a proxy.

rbygrave:
@SentryMan  review the last commit https://github.com/avaje/avaje-inject/pull/896/commits/c0ecf8fe7fd66ec91ea2a38a46c7468be2cfc854 .... that will generate protected methods that throw UnsupportedOperationException like:

```java
  @Override
  protected int next(int arg0) {
    // @Lazy proxy does not support protected methods, instead use @Lazy(useProxy = false)
    throw new UnsupportedOperationException();
  }
```

Without this, user will instead get a compilation error which I think will confuse them so I _think_ this might be better?

SentryMan:
I mean in the first place the generated classes are final so the generated protected methods cannot be accessed

rbygrave:
> the generated protected methods cannot be accessed

Technically they are accessible by classes in the same package as the generated lazy class ... but that does require specific use of the `_$Lazy` type so the _expectation_ is that isn't something people will do.

SentryMan (commented):
Lgtm

SentryMan:
okay good with this

rbygrave (approved):

PR #915, opened 2025-11-09
[events] support custom executor for async, add ObserverManager.Builder

rbygrave:


SentryMan:
works for me

SentryMan (approved):

inject-events/src/main/java/io/avaje/inject/events/spi/ObserverManagerPlugin.java
```diff
@@ -1,18 +1,19 @@
...
   @Override
   public Class<?>[] provides() {
-    return new Class<?>[] {DObserverManager.class};
+    return new Class<?>[] {ObserverManager.class};
   }
 
   @Override
   public void apply(BeanScopeBuilder builder) {
-    builder.provideDefault(null, DObserverManager.class, DObserverManager::new);
+    builder.provideDefault(null, ObserverManager.class, () -> ObserverManager.builder().build());
```
rbygrave: - Existing bug that was registering type DObserverManager.class ...
- Use the builder ...

To specify use of a specific Executor, wire a ObserverManager explicitly via:

```java
@Factory 
class MyFactory {
...

  @Bean
  ObserverManager observerManager(Executor executor) {
    return ObserverManager.builder()
        .asyncExecutor(exec)
        .build();
  }
}
```

... if we need to further customise the ObserverManager, we add more options to the ObserverManager.Builder in the future.

rbygrave:
Oh cool, you were super fast to review ... I was busy explaining this option etc.

Interesting that I think this also fixes a bug there where the plugin wasn't registering with the correct interface ObserverManager.class type there ... (the plugin would create an additional unused DObserverManager instance if folks where overriding providing their own ObserverManager implementation).

PR #916, opened 2025-11-09
error when `io.avaje.inject.spi.AvajeModule` class not found

andponlin:
If the project does not have the `avaje-inject` dependency configured on the project but the code-generator is configured then it will yield an NPE. This fix will at least raise an exception with an explanation.


SentryMan (approved):
One wonders how you got into this situation but the change looks good to me

PR #937, opened 2025-11-27
Fix Lazy Priority Generation

### LINKED ISSUE #936: @Primary ignored when @Lazy present
### Environment
avaje-inject: 12.0
Java: 25

### Single-File Reproducer
SpanishGreeting is marked as @Primary but EnglishGreeting gets injected instead. If I remove @Lazy(Lazy.Kind.PROVIDER) from the classes, SpanishGreeting gets injected correctly.
```java
import io.avaje.inject.BeanScope;
import io.avaje.inject.Lazy;
import io.avaje.inject.Primary;
import jakarta.inject.Singleton;

interface GreetingService {
  String greet();
}

@Lazy(Lazy.Kind.PROVIDER)
@Singleton
class EnglishGreeting implements GreetingService {
  @Override public String greet() { return "Hello"; }
}

@Primary
@Lazy(Lazy.Kind.PROVIDER)
@Singleton
class SpanishGreeting implements GreetingService {
  @Override public String greet() { return "Hola"; }
}

@Lazy(Lazy.Kind.PROVIDER)
@Singleton
class Greeter {
  private final GreetingService service;
  Greeter(GreetingService service) { this.service = service; }
  String run() { return service.greet(); }
}

// Minimal test/runner
public class PrimaryLazyRepro {
  public static void main(String[] args) {
    try (var scope = BeanScope.builder().build()) {
      var greeter = scope.get(Greeter.class);
      String actual = greeter.run();
      System.out.println("Selected: " + actual);
      if (!"Hola".equals(actual)) {
        throw new IllegalStateException("Expected @Primary to be selected, got: " + actual);
      }
    }
  }
}
```

### Question
Is this the intended behavior when using @Lazy, or is it a bug?

SentryMan:
It seems that the `asPrimary` code was not getting generated.

resolves #936

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/SimpleBeanWriter.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/SimpleBeanWriter.java
@@ -178,7 +178,6 @@ final class SimpleBeanWriter {
       indent += "  ";
 
       final String registerProvider;
-
       if (beanReader.proxyLazy()) {
         registerProvider = "registerLazy";
       } else if (beanReader.lazy()) {
--- a/inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/BeanReader.java
@@ -371,7 +371,7 @@ final class BeanReader {
     writer.append("register(bean);").eol();
   }
 
-  public void writePriority(Append writer) {
+  void writePriority(Append writer) {
     if (primary) {
       writer.append("asPrimary().");
     } else if (secondary) {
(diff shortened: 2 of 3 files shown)


inject-generator/src/main/java/io/avaje/inject/generator/SimpleBeanWriter.java
```diff
@@ -186,7 +187,9 @@ private void writeAddFor(MethodReader constructor) {
         registerProvider = "asPrototype().registerProvider";
       }
 
-      writer.append("      builder.%s(() -> {", registerProvider).eol();
```
rbygrave: bug here, missing priorities ...

PR #953, opened 2025-12-10
Make `isBeanAbsent` not count secondary beans

### LINKED ISSUE #952: No nima controllers with avaje-http-inject-plugin version 3.6, which breaks isBeanAbsent() check
### Current Workaround 

The current workaround to this issue is to `avaje-http-inject-plugin` version `3.5`

----------

So bit of an interesting issue. Bumping up a nima app to http 3.6, includes the `avaje-http-inject-plugin` version 3.6, which registers an error plugin via:

```
 builder.provideDefault(HttpFeature.class, HelidonHandler::new);
```
... and the little detail here is that this effectively registers a component for `HttpFeature.class` on effectively the **_parent_** BeanScope [and this then effectively breaks the `isBeanAbsent()` check].

### Background:
- The `isBeanAbsent()` check is designed to check for test doubles and supplied beans and to skip registering the 'real' component if a test double has already been supplied. That is, it's for testing purposes.
- It's got a kinda design flaw where ... these types passed in to `isBeanAbsent()` are also the types that the bean will register against [so these types have almost 2 uses but they are not treated separately]. 
- We desire to register controllers against the `HttpFeature.class` type [so that we can register all of them with Helidon later], but we don't ever really desire to use that type to check for a test double.
- This all worked as long as there wasn't any `HttpFeature.class` type registered on the parent BeanScope [but the error handler plugin changed that situation].

### Options to resolve this

1. Modify DBuilder.removeAnnotations() ... to also remove `HttpFeature.class` (and similar Jex type Routing.HttpService.class and similar Javalin type). This method removes any types that we do not wish to consider for `isBeanAbsent()`.
 
2. Add some method like `registerTypes()` for types for a component to register against but that are not considered for "test double check". Currently all the types go to `isBeanAbsent()` but instead some types could go to a new `registerTypes()` method.

3. Entirely skip the `isBeanAbsent()` for the generated Router components. That is, the actual generated http routers are never really candidates for "test doubles" anyway. Adjust the generated code to not use `isBeanAbsent()` and replace that with a new `registerTypes()` method.

4. Can we use `@BeanTypes` to resolve this? Not yet, but if we split the list of types into 2 lists where one is for `isBeanAbsent()` and the other for `registerTypes()` then ... maybe that would work.

--------

Example generated router DI code:

```java
public final class UserController$Route$DI  {

  public static final Type TYPE_SupplierHttpFeature =
      new GenericType<Supplier<HttpFeature>>(){}.type();

    if (builder.isBeanAbsent(UserController$Route.class, HttpFeature.class, TYPE_SupplierHttpFeature, Supplier.class, ServerLifecycle.class)) {
      var bean = new UserController$Route(builder.get(UserController.class,"!controller"), builder.get(Jsonb.class,"!jsonb"));
      builder.register(bean);
    }
  }

}

```

So if we split the types it could be something like:

```java
// only check UserController$Route.class for a test double
if (builder.isBeanAbsent(UserController$Route.class)) { 
  // register extra types but NOT check these types for test doubles
  builder.registerTypes(HttpFeature.class, TYPE_SupplierHttpFeature, Supplier.class, ServerLifecycle.class);
  var bean = new UserController$Route(builder.get(UserController.class,"!controller"), builder.get(Jsonb.class,"!jsonb"));
  builder.register(bean);
}
```
... which is almost like modifying `@BeanTypes` and putting it on the generated routers like:

```java
@Generated("avaje-helidon-generator")
@Component
@BeanTypes(
  value = UserController$Route.class, // types to check for test doubles
  registerTypes=HttpFeature.class)     // extra types to register
public final class UserController$Route implements HttpFeature {
```
... and then the generated code would be:

```java
if (builder.isBeanAbsent(UserController$Route.class)) { 
  builder.registerTypes(HttpFeature.class);
  var bean = new UserController$Route(builder.get(UserController.class,"!controller"), builder.get(Jsonb.class,"!jsonb"));
  builder.register(bean);
}
```

Hmm. Thoughts?

SentryMan:
Secondary beans should not count if there are alternatives.

resolves #952

### DIFF
--- a/inject/src/main/java/io/avaje/inject/spi/DBeanScope.java
+++ b/inject/src/main/java/io/avaje/inject/spi/DBeanScope.java
@@ -179,10 +179,10 @@ final class DBeanScope implements BeanScope {
   }
 
   static <T> List<T> combine(List<T> values, List<T> parentValues) {
-    if (values.isEmpty()){
+    if (values.isEmpty()) {
       return parentValues;
     }
-    if (parentValues.isEmpty()){
+    if (parentValues.isEmpty()) {
       return values;
     }
     values.addAll(parentValues);
--- a/inject/src/main/java/io/avaje/inject/spi/DContextEntry.java
+++ b/inject/src/main/java/io/avaje/inject/spi/DContextEntry.java
@@ -40,16 +40,11 @@ final class DContextEntry {
 
   /** Check if a non secondary entry exists */
   Object nonDefaultEntry(String name) {
-    DContextEntryBean entry;
     if (entries.size() == 1) {
-      entry = entries.get(0);
-      return entry.isNameMatch(name) && entry.priority() != BeanEntry.SECONDARY
-          ? entry.bean()
-          : null;
+      return entries.get(0).nonDefaultMatch(name);
     }
-    entry = new EntryMatcher(name, null).findMatch(entries);
-
-    return entry != null && entry.priority() != BeanEntry.SECONDARY ? entry.bean() : null;
+    var entry = new EntryMatcher(name, null).findMatch(entries);
+    return entry != null ? entry.nonDefaultMatch(null) : null;
   }
 
   Object get(String name, Class<? extends AvajeModule> currentModule) {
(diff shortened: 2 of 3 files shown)


rbygrave (approved):

rbygrave:
BTW: Thanks for seeing and providing this solution. I went off down the types track there and didn't see this option.

PR #954, opened 2025-12-10
Add BeanTypes registerTypes, for separate set of types that are excluded from isBeanAbsent()

rbygrave:
These register types are not included in the isBeanAbsent() check

-----------

Currently the thinking is to NOT merge this PR, in that it isn't strictly required and possibly will not be.


rbygrave:
Closing this. We are NOT expecting to ever need it at this stage.

PR #955, opened 2025-12-10
Add avaje-inject-bom bill of materials

rbygrave:


SentryMan:
perhaps we should make a new repo with a general avaje bom

PR #1025, opened 2026-05-08
Add an Annotation to Register the Main Class

SentryMan:
So there I am, putting the main class in my jar plugin once again, then it hits me. "Why not have the generator do this for me?" 

- adds a new annotation to detect the main class
- generator will read the annotation then update the pom.xml to add/update the maven jar plugin
- for gradle, generates a file read by the inject gradle plugin to set the main class in the jar tasks
- activates strict wiring by default, with the option of interweave


stevenschlansker:
Totally unsolicited opinion, I think `@MainClass` is a fine name. It's direct and accurate, and I don't see any collisions with other class names in the classpath of the first random project I looked at.
My other thought was `@Entrypoint` but I don't think that's better.

rbygrave (approved):

PR #1027, opened 2026-05-08
Ensure Closeable beans are closed when BeanScope creation fails

### LINKED ISSUE #1026: [BUG] Beans in a scope not closed if scope construction or postconstruct fails
## Describe the Bug

avaje-inject 12.5
java 26

When developing test cases, sometimes you break the module under test in such a way that wiring up BeanScope fails during construction or during PostConstruct. In this case, Closeable beans that have already been constructed are never closed - they're "leaked". This can lead to misleading test failures: if your fixture allocates a scarce resource (say, a database connection) which it expects to cleanup in `close()`, later fixtures then start to fail 'Couldn't acquire a database connection' rather than whatever your actual test failure is.

## Minimal Reproducible Example

```java
@Factory
@Profile("partial-build-failure-test")
public class FailingFactory {

  public static final AtomicBoolean FIRST_CLOSED = new AtomicBoolean();

  @Bean
  First first() {
    return new First();
  }

  @Bean
  String second(First f) {
    throw new IllegalStateException("boom while wiring String from " + f);
  }

  public static final class First implements AutoCloseable {
    @Override
    public void close() {
      FIRST_CLOSED.set(true);
    }
  }
}

class PartialBuildFailureTest {

  /**
   * When a factory method throws partway through wiring, AutoCloseable beans
   * that were already constructed and registered must still be closed.
   * Otherwise they leak resources (file handles, sockets, threads, etc.).
   */
  @Test
  void buildFailureClosesAlreadyConstructedAutoCloseables() {
    FailingFactory.FIRST_CLOSED.set(false);

    assertThatThrownBy(() -> BeanScope.builder().profiles("partial-build-failure-test").build())
        .hasMessageContaining("boom");

    assertThat(FailingFactory.FIRST_CLOSED)
        .as("AutoCloseable beans constructed before the failing factory must be closed when the build aborts")
        .isTrue();
  }
}
```

Similar setup for PostConstruct.

## Expected Behavior

When a BeanScope fails construction due to a constructor inject failure, or a PostConstruct failure, all created AutoCloseable beans should be closed to release any OS resources (db connections, open files).

## Actual Behavior

The beans are not closed, and OS resources they hold leak.

stevenschlansker:
This is an AI assisted attempt to fix issue #1026

The test cases included highlight that when a BeanScope fails to create due to a constructor injection throwing, or a PostConstruct throwing, beans' close() methods are not called.

I have no particular attachment to this code -- please feel free to rip this draft apart for parts, or ignore it, or modify it as you like -- Claude doesn't care ;)

### DIFF
--- a/blackbox-test-inject/src/main/java/org/example/myapp/partialbuild/FailingFactory.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/partialbuild/FailingFactory.java
@@ -1,0 +1,35 @@
+package org.example.myapp.partialbuild;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+import io.avaje.inject.Bean;
+import io.avaje.inject.Factory;
+import io.avaje.inject.Profile;
+
+/**
+ * Test fixture for verifying that AutoCloseable beans are closed when a
+ * {@code @Factory} method throws partway through wiring.
+ */
+@Factory
+@Profile("partial-build-failure-test")
+public class FailingFactory {
+
+  public static final AtomicBoolean FIRST_CLOSED = new AtomicBoolean();
+
+  @Bean
+  First first() {
+    return new First();
+  }
+
+  @Bean
+  String second(First f) {
+    throw new IllegalStateException("boom while wiring String from " + f);
+  }
+
+  public static final class First implements AutoCloseable {
+    @Override
+    public void close() {
+      FIRST_CLOSED.set(true);
+    }
+  }
+}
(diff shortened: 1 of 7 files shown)


SentryMan:
perhaps instead of a static class you could expose a close method to builder

stevenschlansker:
OK, happy to make that update, one moment...

stevenschlansker:
thanks for taking a look @SentryMan , I updated it to add a new public `closeOnFailure` method instead.
I elected not to use Closeable because then static checkers will complain everywhere that you're not closing your builder, which is normal.

SentryMan (approved):
works for me

rbygrave (approved):

PR #1028, opened 2026-05-08
FactoryOrder: preserve module order when supplied beans satisfy a requires

stevenschlansker:
This is an AI-assisted attempt to fix a purported bug in module ordering. Human-me had already run into confusing cases regarding module ordering with @External beans, without really understanding what was wrong. Claude's analysis, claiming this is a bug or shortcoming in Avaje, follows. I've attempted to verify it myself but some extra scrutiny might be warranted before accepting this change. Same as before - feel free to totally junk the code if you don't find it acceptable; happy to write this up as simply a bug report with test if that's preferred:

When BeanScopeBuilder.build() is called with one or more supplied beans, FactoryOrder runs with suppliedBeans=true. If the strict topological pass stalls, the existing fallback dumps the remaining queue in insertion order, discarding the partial order it had already computed.

This breaks any module whose @InjectModule(requires=...) names a type that is supplied via BeanScopeBuilder.bean() (or only partially provided through a multi-provider chain). With modules added in [B, A] where A provides TypeX (consumed by B) and A's other requires are externally supplied, B gets wired before A and Avaje throws "Injecting null for TypeX ... potential beans to inject: []" at runtime.

Add a relaxed second pass that runs only when suppliedBeans=true and only after the strict pass has stalled:
  - a required type with no module-level provider is treated as externally supplied (assume it came from .bean());
  - a required type counts as satisfied once at least one of its providers has been pushed.

Anything still queued after the relaxed pass has an in-module provider that was never pushed (a cycle or otherwise unresolvable case); those still emit in insertion order, as before. Existing behavior for the strict path is unchanged.

### DIFF
--- a/inject/src/main/java/io/avaje/inject/DBeanScopeBuilder.java
+++ b/inject/src/main/java/io/avaje/inject/DBeanScopeBuilder.java
@@ -388,12 +388,21 @@ final class DBeanScopeBuilder implements BeanScopeBuilder.ForTesting {
     private void processQueue() {
       int count;
       do {
-        count = processQueuedFactories();
+        count = processQueuedFactories(false);
       } while (count > 0);
 
-      if (suppliedBeans) {
-        // just push everything left assuming supplied beans
-        // will satisfy the required dependencies
+      if (suppliedBeans && !queue.isEmpty()) {
+        // Relaxed pass: when supplied beans may satisfy unknown dependencies,
+        // treat a required type with no module-level provider as externally
+        // supplied, and treat a required type with at least one provider already
+        // pushed as satisfied. This preserves topological order between modules
+        // whose only unsatisfied dependency is supplied via BeanScopeBuilder.bean()
+        // or comes from a partial multi-provider chain.
+        do {
+          count = processQueuedFactories(true);
+        } while (count > 0);
+        // Anything still queued has an in-module provider that was never pushed
+        // (a cycle or otherwise unresolvable case); emit in insertion order.
         for (final FactoryState factoryState : queue) {
           push(factoryState);
         }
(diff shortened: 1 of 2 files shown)


SentryMan:
you're going to have to describe the bug in more concrete terms

stevenschlansker:
Did you take a look at the test case? I think it's a pretty good explainer. I'll try to write it out in words:

Given: a module B that provides nothing and requires X, and a module A that provides X and requires @External Y
Register modules B, then A (ordering important), the only correct ordering is A, B (due to X dependency) but currently avaje-inject tries to wire B, A (unmet external dependency Y short-circuits module ordering) which then fails due to a missing dependency X.

SentryMan:
I think the new interweave thing we added before might resolve your issue

stevenschlansker:
Oh! I'm happy to try that out - I'll grab the latest snapshot and re-test. Is there a specific way I need to turn on interweave or is it always on?

SentryMan:
you do `@InjectModule(interweave=true)`

stevenschlansker:
Is there a reason not to just turn it on everywhere? i.e., why is this not the default / when should I *not* use interweave?

stevenschlansker:
Oh, looks like you can't have more than a single module marked as interweave... I think that might be a deal-breaker for us, since this problem crops up in multiple places in our codebase. I'll continue investigating, but my 2-minute analysis says that interweave is probably not a total fix for us.

SentryMan:
It should only be used on the project with the main class. It essentially locks in the module ordering for all modules at compile time. So if the project is used as a library for another avaje project the one with interweave will override

stevenschlansker:
I'll spend some more time playing with interweave and see if I can't get it working for us, or try to come back with a more convincing "it isn't a sufficient fix because of reason X". Thanks for your continued assistance and patience with my rapid-fire bug reporting :)

SentryMan:
So like if there are external modules, interweave will essentially call their methods from the generated module
```java

@InjectModule()
@SuppressWarnings("all")
public final class MyappModule implements AvajeModule {

  /**
   * Creates all the beans in order based on constructor dependencies.
   * The beans are registered into the builder along with callbacks for
   * field/method injection, and lifecycle support.
   */
  @Override
  public void build(Builder builder) {
    // create beans in order based on constructor dependencies
    // i.e. "provides" followed by "dependsOn"
    build_pkg_private_AdderImpl(builder);
    build_myapp_ListFactory(builder);
    build_interweave_LocalImpl(builder);
    //..
    //calling external modules
    OneModule.build_one_OtherComponent(builder);
    OneModule.build_one_OtherComponentGeneric(builder);
    OneModule.build_interweave_BeanFromOther(builder);
    ExternalModule.build_custom_OtherCustomComponent(builder);
    ExampleExternalAspectModule.build_sub_MyAspectImplementation(builder);
   //..
   //other build methods
  }
```

rbygrave:
We need to remember to "Respect the bug" and not just lean on a workaround to the issue.

It is nice having workarounds, it is actually better to fix a underlying bug though.  Being explicit, the interweave feature is a _workaround_ and not a _bug fix_.

stevenschlansker:
I think interweave is not a good fit for our project, unfortunately. It seems like a great feature for "closed world" builds but we ship (simplified) a platform library and a bundle of services, and we expect to link their client libraries together dynamically.

For example, we might ship
lib 1.0 + service A 1.0
lib 1.1 + service A 1.1
lib 1.0 + service B 1.0

and then expect to dynamically link service A 1.1 and B 1.0 together (integration testing).
This change of lib 1.0 -> 1.1 at runtime behind service B's back seems to be incompatible with interweave, since it peeks through the "public avaje module api" to individual dependencies which might be re-wired in a patch release.

Please let me know if I've misunderstood.

I think the unit test included in this PR is a decent regression test, but it doesn't actually show the shape of the user code that fails. So, I had my assistant draft up an example project.

This project contains two modules that provide / depend on a dependency, and an @External dependency that messes up the module ordering.

This test fails against `master` branch, but succeeds with this patch applied. Let me know what you think.
https://github.com/stevenschlansker/avaje-inject/commit/e596d998d98fab8a57568eb02aba05fe8831f416

SentryMan:
I see, now this I can work with

SentryMan:
Still working on it, but you can try strict wiring instead of interweave in the meantime. It is similar in that it also locks module order, but will not have a hard dependency on the public build methods.

SentryMan (approved):
Yeah I can't think of anything better

cbarlin:
Unsolicited, but should this also apply when generating with `strictWiring = true` (so the `inject-generator` would need to be able to do the same thing)?

As in "sees" that there are external beans, and does it's best to order without those?

SentryMan:
This bug cannot happen when strict wiring is on, as the module order is fixed and cannot be changed.

rbygrave (approved):

SentryMan:
12.6-RC3

PR #1031, opened 2026-05-19
Fix strict wiring discovery bug

### LINKED ISSUE #1030: [BUG] `strictWiring=true` writes in discovery order, not correct order
## Describe the Bug

When using `strictWiring = true` on an application with a number of dependency modules, the written out `CompiledOrder` contains the modules in discovery order, rather than the order needed to correctly start the application. Without `strictWiring`, the application works as intended.

## Environment Info
Avaje Inject 12.5 (also on 12.6-RC2)
Java 25

- [x] Maven
- [ ] Gradle

Plugin is being used, as are a number of Avaje modules (Jex for HTTP, HTTP Client, Jsonb, Validator, Configuration)

## Expected Behaviour

When run, the application starts and behaves correctly (same as it does without strict wiring)

## Actual Behaviour

When run, the application fails to start because the order in which it tries to load the modules is incorrect, causing it to fail with an `Injecting null for ...` error.

## Minimal Reproducer

A "minimal" reproducer would not be that small, and the code base I'm experiencing this on is proprietary. I could try and create one if it's really needed but I've tried to trace where the issue is occurring below.

## Additional Context

Stepping through the generator when it's working out the module order, it seems that the `ProcessingContext.externallyProvided(dependency)` check in [`notProvided` in `FactoryOrder`](https://github.com/avaje/avaje-inject/blob/d3d6d298033e4844936fa0c44780811cdd8ff820/inject-generator/src/main/java/io/avaje/inject/generator/FactoryOrder.java#L114) seems to always mark the dependency as "satisfied" when it isn't (yet).

The Maven plugin does seem to be correctly generating the `avaje-module-dependencies.csv` correctly. Stepping through, it looks like the generator is reading that CSV and then marking everything it reads in as "provided" in [`init` in `InjectProcessor`](https://github.com/avaje/avaje-inject/blob/d3d6d298033e4844936fa0c44780811cdd8ff820/inject-generator/src/main/java/io/avaje/inject/generator/InjectProcessor.java#L89), which it does before then working out the module order.

There are no external beans, and no additional scopes - `BeanScope.builder().build()` picks up everything.

SentryMan:
Resolves #1030 by skipping the externallyProvided path for dependencies in providesMap

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/FactoryOrder.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/FactoryOrder.java
@@ -109,8 +109,11 @@ class FactoryOrder {
 
   private boolean notProvided(String dependency) {
     final FactoryList factoryList = providesMap.get(dependency);
-    return (factoryList == null || !factoryList.allPushed())
-      && !pluginProvided.contains(dependency)
+    if (factoryList != null) {
+      // dependency is owned by a known module, only satisfied once that module is pushed
+      return !factoryList.allPushed();
+    }
+    return !pluginProvided.contains(dependency)
       && !ProcessingContext.externallyProvided(dependency);
   }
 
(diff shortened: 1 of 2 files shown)


cbarlin:
It's seriously impressive how quickly you can turn around a fix 😅 

Great work man, thanks heaps!

rob-bygrave (approved):

PR #1050, opened 2026-08-05
Fix Self-Providing Beans

### LINKED ISSUE #1049: Bug: Wiring order incorrect when @Profile excludes one of multiple interface implementations
**Version:** 12.6

**Description:**
When two beans implement the same interface and the first is excluded via 
`@Profile`, the second implementation is instantiated AFTER the service 
that depends on the interface, causing a wiring failure.

**To reproduce:**
```
@Singleton
@Profile("cloud")
public class CloudService implements MyService { }

@Singleton
@Profile(none = "cloud")
public class LocalService implements MyService { }

@Singleton
public class Consumer {
    @Inject MyService service;  // fails — LocalService not yet instantiated
}
```

**Expected:** When CloudService is excluded by profile, LocalService should 
be instantiated before Consumer.

**Actual:** The compile-time generated wiring order assumes CloudService 
satisfies the dependency. At runtime when it's excluded, LocalService 
is still ordered after Consumer.

**Workaround:** Using `Provider<MyService>` to defer resolution.

### Additional findings:

1. **`@RequiresBean` does NOT fix the ordering either.** Using 
   `@RequiresBeanMissing(CloudService.class)` on `LocalService` still 
   results in `LocalService` being ordered after `Consumer`. The issue 
   is not about selection/conditionality — it's about wiring order.

2. **`@Secondary` / `@Primary` also do not help** — they affect 
   injection priority, not instantiation order.

3. **Root cause clarification:** The generator only delays consumers 
   for the *first* bean it resolves as the interface provider. When 
   that bean is excluded by `@Profile` at runtime, other 
   implementations of the same interface retain their original 
   (later) position. 
   
   **The fix should ensure ALL beans providing a given interface type 
   are ordered before ANY consumer of that interface**, not just the 
   one the compile-time generator happened to resolve first.

4. The `Provider<T>` workaround works but is unnecessary indirection 
   for `@Singleton` beans — the real fix belongs in the generator's 
   `shouldDelay` logic or equivalent ordering mechanism.

SentryMan:
Resolves #1049 

  Profile was a red herring,  In the issue reproducer, `ConnectionPool` implemented `ConnectionFactory`, so
  factory method `b(ConnectionFactory a) ` was registered as a provider of `ConnectionFactory` (its own dependency) which messed up the ordering. 

- exclude the bean itself from its own provider-list check.

### DIFF
--- a/blackbox-test-inject/src/main/java/org/example/myapp/profileorder/Conn.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/profileorder/Conn.java
@@ -1,0 +1,3 @@
+package org.example.myapp.profileorder;
+
+public interface Conn {}
--- a/blackbox-test-inject/src/main/java/org/example/myapp/profileorder/DbService.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/profileorder/DbService.java
@@ -1,0 +1,16 @@
+package org.example.myapp.profileorder;
+
+import io.avaje.inject.Profile;
+import jakarta.inject.Singleton;
+
+@Singleton
+@Profile("cloud")
+public class DbService implements MyService {
+
+  public DbService(Pool pool) {}
+
+  @Override
+  public String name() {
+    return "db";
+  }
+}
--- a/blackbox-test-inject/src/main/java/org/example/myapp/profileorder/FileService.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/profileorder/FileService.java
@@ -1,0 +1,14 @@
+package org.example.myapp.profileorder;
+
+import io.avaje.inject.Profile;
+import jakarta.inject.Singleton;
+
+@Singleton
+@Profile(none = "cloud")
+public class FileService implements MyService {
+
+  @Override
+  public String name() {
+    return "file";
+  }
+}
--- a/blackbox-test-inject/src/main/java/org/example/myapp/profileorder/MyService.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/profileorder/MyService.java
@@ -1,0 +1,6 @@
+package org.example.myapp.profileorder;
+
+public interface MyService {
+
+  String name();
+}
--- a/blackbox-test-inject/src/main/java/org/example/myapp/profileorder/Pool.java
+++ b/blackbox-test-inject/src/main/java/org/example/myapp/profileorder/Pool.java
@@ -1,0 +1,4 @@
+package org.example.myapp.profileorder;
+
+/** Mimics r2dbc ConnectionPool which itself implements ConnectionFactory. */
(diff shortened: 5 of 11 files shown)


rbygrave:
Rob: Should we copy over the test from #1051 into here? ... and look to confirm the gap for the "self-only edge case" and fix?

Clanker review:
---------
Review of PR #1050

Same root cause diagnosis as mine (self-providing bean like  ConnectionPool implements ConnectionFactory  forces the "last ditch" round). Their fix is structurally similar — exclude  self  from the provider-list check — but with a different safeguard for the "only provider is self" case:

```java
private boolean isWired(boolean anyWired, MetaData self) {
  if (containsOnly(self)) {
    return false;   // hard-coded false, doesn't check self.isWired()
  }
  ...
}
```

## Bug found:  
containsOnly  checks  list.size() == 1 , but the provider map registers each bean under multiple keys ( toString() ,  type() , each  provides  entry) — so for a bean depending on its own exact type (a genuine direct self-dependency, e.g.  class Foo { Foo(Foo f) } ), the provider list for that type has 2 entries (added once via  providerAdd(type()) , once via  providerAdd(toString()) ), both pointing to the same  MetaData  instance.  containsOnly  ( list.size() == 1 ) never triggers, so it falls through to  isAllWired / isAnyWired , which skip  self  in the loop, find nothing left to check, and return  true  — silently marking a genuinely-circular self-dependent bean as wired, with no compile error reported. I verified this concretely:

remaining=0 isWired=true   // should be a compile error, not silently satisfied

My fix avoids this because it checks "does any other distinct MetaData exist" via iteration ( hasOther ) rather than  list.size() == 1 , so duplicate self-entries under different keys don't fool it — it correctly falls back to strict behavior and leaves the bean unwired/reported.

## Recommendation: 
worth commenting on PR #1050 with this finding — a bean with a direct/circular self-dependency and no other implementation would silently compile instead of failing with a clear circular-dependency error. Their blackbox test additions ( ProfileOrderTest ) are good and pass fine; _the gap is specifically in the self-only edge case_, which their own test suite doesn't cover.

SentryMan:
very well

rbygrave:
Yup, looks good. I'll approve.

> very well

Just to say I'm generally unclear how to interpret "very well" ... but I tend to interpret it as "I'm not sure if I really agree with you here Rob but I'll do what is asked anyway". I'd say that if "your spidey sense" has gone off then I'd encourage something else even if its just "Hmm, not sure - seems marginal, let me think about that ..."

I'm going with "very well" meaning "Yup, all good - done" here and that you didn't have any regrets with the last commit.

rbygrave (approved):

SentryMan:
> I'm going with "very well" meaning "Yup, all good - done" here 

yeah that's what I meant here

PR #1051, opened 2026-08-05
Fix for #1049 Wiring ordering issue with a bean with self dependency

rbygrave:
Root cause: MetaDataOrdering's "last ditch" Round C relaxes "all providers wired" to "any provider wired" — but as a global flag, not scoped to the specific interface that needs it. A decorator/wrapper bean (e.g. ConnectionPool implements ConnectionFactory while depending on a ConnectionFactory) is a member of its own provider list and can never be "all wired" before itself, so it permanently forces the entire remaining queue into Round C. That lets an unrelated Consumer wire via just one of two @Profile-conditional MyService impls before the other is ready — breaking ordering guarantees.

Fix: ProviderList.isAllWired/isAnyWired now exclude the bean itself from its own provider-list check (with a safeguard: if it's the only provider, fall back to strict behavior so genuine circular self-dependencies still fail correctly). This resolves decorator beans in the normal strict rounds, avoiding the Round C leniency entirely for this pattern.

Validation:

• New Issue1049Test.java: real-world reproducer test (mirroring ConnectionFactory/ConnectionPool) passes; self-dependency safeguard test passes; one synthetic test documenting a distinct, broader, still-open edge case is @Disabled with explanation

### DIFF
--- a/inject-generator/src/main/java/io/avaje/inject/generator/MetaDataOrdering.java
+++ b/inject-generator/src/main/java/io/avaje/inject/generator/MetaDataOrdering.java
@@ -169,7 +169,10 @@ final class MetaDataOrdering {
     String dependencyName = dependency.name();
     var providerList = providers.get(dependencyName);
     if (providerList != null) {
-      return providerList.isWired(anyWired);
+      // exclude queuedMeta itself - a bean that both depends on and provides (e.g. implements)
+      // the same type (a "decorator"/wrapper style bean) can never be wired before itself, so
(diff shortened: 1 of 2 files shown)


SentryMan:
which would you rather?  this or the PR I made before this one?

SentryMan (changes requested):
missing blackbox tests

rbygrave:
Replaced by https://github.com/avaje/avaje-inject/pull/1050

PR #1052, opened 2026-08-05
 Order Providers Ahead of Consumers in the Last Ditch Wiring Round

SentryMan:
Follow-up to #1050, closing the gap Rob flagged there. 

The last ditch round now collects everything it makes ready, then orders that set so any bean providing a dependency of another comes first (`providersFirst`, a DFS over the existing `resolveQueuedDeps` provider graph). Beans in a cycle keep their original relative order.


rbygrave (approved):
Awesome !!

PR #1056, opened 2026-08-14
AutoProvidesMojo: warn and skip unresolvable InjectExtension entries

stevenschlansker:
Since the generated services files also list the modules of upstream dependencies, an entry can name a class that is not resolvable on the compile-scope classpath (an optional, test-scoped, or excluded dependency). The ServiceLoader iteration then threw an uncaught ServiceConfigurationError that named neither the declaring jar nor the reason the class was unresolvable, and failed the build.

Iterate the ServiceLoader stream catching the per-element error, which skips just the unloadable entry; such a module is not a compile-time concern for the project being built. A separate scan of the services resources is used only to name, in the warning, the services file that declares each skipped entry. The extensions are loaded once and shared by the plugin and module CSV writers.

### DIFF
--- a/inject-gradle-plugin/build.gradle
+++ b/inject-gradle-plugin/build.gradle
@@ -18,6 +18,7 @@ dependencies {
   implementation gradleApi()
 
   testImplementation 'org.junit.jupiter:junit-jupiter-api:5.8.1'
+  testImplementation 'org.assertj:assertj-core:3.27.6'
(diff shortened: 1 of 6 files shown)


SentryMan:
@stevenschlansker I made some changes, let me know if you're alright with them

rbygrave (approved):

stevenschlansker:
Thanks!!

PR #1061, opened 2026-08-14
Delay Soft Dependency module ordering

### LINKED ISSUE #1060: [BUG] List injection can lose elements depending on classpath order
## Describe the Bug

List injection element resolution depends on classpath ordering due to a limitation in the dependency solver.
Blackbox example includes a M1 that declares an extension interface and a M2 that implements it. Injecting `List<Contrib>` does not include M2's contributed bean if M1 comes before it in ServiceLoader iteration order (classpath order).

## Environment Info
avaje-inject 12.7-RC2 + 12.6
java 26 maven

## Minimal Reproducible Example

Module m2 contributes an implementation of `Contrib` (an interface defined in a shared api module):

```java
@Singleton
public class NamedContrib implements Contrib {}
```

Module m1 consumes it via `List` injection, with no `Contrib` provider visible at compile time:

```java
@Factory
public class WireFactory {
  
  @Bean
  Mapper mapper(List<Contrib> contribs) {
    return new Mapper(contribs); // contribs is empty when m1 wires before m2
  }
}
``` 

The generated `M1Module` has no `requiresBeans()` entry; the dependency appears only as `soft:...api.Contrib` in bean-level `@DependencyMeta`, so module wiring has no edge between m1 and m2 and falls back to classpath order in FactoryOrder. Full runnable reproduction in #1059.

## Expected Behavior

List injection should resolve both components regardless of classpath ordering M1 vs M2
Soft dependencies could be treated as "order if present" constraints in the dependency solver.

## Actual Behavior

With M1 before M2, the contributed extension is lost with no warning or error.

### LINKED ISSUE #1066: [BUG] Conditional beans with in-module dependencies misses necessary ordering edge
## Describe the Bug

Given M1 and M2, M1 provides a `@RequiresProperty(value = "greet.mode", equalTo = "app") AppGreeting implements Greeting` and a `Reader` which consumes `Greeting`. M2 provides `@RequiresProperty(value = "greet.mode", equalTo = "alt") AltGreeting implements Greeting`.

With `greet.mode=alt`, wiring fails if M1 comes before M2 (`.modules(new M1Module(), new M2Module())`, or depending on classpath order)
With `greet.mode=app`, wiring succeeds in any order.

The generated M1Module has no softRequires (or requires) entry on Greeting because the in-module provider satisfies `MetaDataOrdering.dependencySatisfied` regardless of the condition.

Workaround: `@InjectModule(requires = Greeting.class)` works around the discovery issue.

The current open #1061 does not fix this issue - it is separate (verified it does not fix provided test)

## Environment Info
avaje-inject 12.6, 12.7-RC2
java 26 maven

## Minimal Reproducible Example

Please see 
https://github.com/avaje/avaje-inject/pull/1064

## Expected Behavior

Modules M1, M2 should wire the same as modules M2, M1

## Actual Behavior

Depending on module wiring order, no bean is found - M1,M2 -> fatal error.

```
IllegalStateException: Injecting null for org.example.condorder.api.Greeting
  name:!greeting when creating class org.example.condorder.m1.Reader
  - potential beans to inject: []. Check for missing module? ...
```

## Additional Context

Possible fix proposed as https://github.com/avaje/avaje-inject/pull/1065

SentryMan:
- Adds a new `softRequiresBeans` method to `AvajeModule`
- updates generation and factory ordering to use this attribute for ordering
- soft beans will delay ordering when possible
- Adds Cross-Collection Blackbox Modules
- change javax scripts to auto handle new maven modules 

Resolves #1060 
Resolves #1064 
Resolves #1066


SentryMan:
@rbygrave have you anything to add here about this approach? The alternative is to do nothing and have `@PostConstruct` be the way of handling this situation.

stevenschlansker:
> have @PostConstruct be the way of handling this situation.

Thanks for working on this fix. If we choose the PostConstruct road, I want to highlight an incomplete fix: PostConstruct cannot correctly handle `@RequiresBean` conditioned definitions. The requirement is evaluated at module build time, and is permanent: once it observes a bean as absent, it is never reconsidered, even if the bean becomes present later during initialization. Correctly sorting the soft dependencies fixes this class of error (as this PR does, I verified); it cannot be worked around with PostConstruct.

rbygrave:
@SentryMan  @stevenschlansker  please review the last commit https://github.com/avaje/avaje-inject/pull/1061/changes/ad2d5d45d66c0703b3d5955b72fecfa1442393be ... and confirm you are happy with it. Thanks.

SentryMan:
works for me

rob-bygrave:
I think we can merge this in. @stevenschlansker if you have concerns on the last commit let on this PR let us know.

rob-bygrave (approved):

stevenschlansker:
Thank you so much for all the fixes! I'm looking forward to our next migration iteration, I feel like we're close now!

SentryMan:
Would you want an rc now or else wait for the other issues to be resolved first

PR #1063, opened 2026-08-17
Add the Ability to Pick Test Modules

### LINKED ISSUE #1058: [Enhancement] module selection for the avaje-inject-test global scope
**Is your feature request related to a problem? Please describe.**

The global test scope built by avaje-inject-test wires every `AvajeModule` on the test classpath. `@InjectTest` offers only `profiles` and `scopePerTest`, so there is no way to bound the wiring. In a multi-module project adopting avaje-inject incrementally, any module whose beans cannot be built in a test context (needs a database, a port, external config) breaks the tests of every module downstream of it, forcing all-or-nothing conversion.

**Describe the solution you'd like**

An opt-in way to bound the wiring, for example `@InjectTest(modules = {UtilModule.class})`, mirroring what `BeanScope.builder().modules(...)` already allows in main code.

Closest prior issue: #7 (2018) asked for the same against the old BootContext API and was closed when that API changed.

SentryMan:
Resolves #1058 by:

1. Using the Serviceloader and filtering for the provided module
2. Using reflecting to instantiate the module if it was not registered as a service
3. Making the global all beans test scope lazy (only load the whole classpath if a test does not specify modules)


inject-test/src/main/java/io/avaje/inject/test/GlobalTestBeans.java
```diff
@@ -95,13 +104,32 @@ void close() {
...
       return allBeans;
     }
 
+    private void initAllBeans() {
+      if (allBeans == null) {
+        allBeansLock.lock();
+        try {
+          if (allBeans == null) {
+            log.log(DEBUG, "Wiring all beans for the test BeanScope");
+            allBeans = BeanScope.builder()
```
stevenschlansker: This looks like double-checked locking, but varies from the "safe" pattern: `allBeans` and `plugin` are not marked volatile, and including both fields in the double-checked lock introduces new possible races.

For example, since `allBeans` is used as the guard value, it is possible a competing thread will see `allBeans` initialized but not see `plugin` being set.

I recommend fixing this one of three ways. Simpler option: remove double checked locking entirely, the only cost is an extra lock each time you inspect the beans, until or unless an actual performance bottleneck is observed.

Or, it is possible to correct the DCL pattern in one of two ways.
Choice 1: introduce a wrapper class holding `plugin` and `allBeans` as `final` and hold a reference to that (classic single-var DCL).
Choice 2: both `plugin` and `allBeans` must be `volatile`, and `allBeans` field must be set *after* the `plugin` field is set, to avoid a racer that sees one field set but not both.

inject-test/src/main/java/io/avaje/inject/test/MetaInfo.java
```diff
@@ -4,16 +4,25 @@
...
 
+  @SuppressWarnings("unchecked")
+  private static final Class<? extends AvajeModule>[] NO_MODULES = new Class[0];
+
   private final MetaReader reader;
+  private final Class<? extends AvajeModule>[] modules;
 
   MetaInfo(Class<?> testClass, Plugin plugin) {
     this.reader = new MetaReader(testClass, plugin);
+    final InjectTest injectTest = testClass.getAnnotation(InjectTest.class);
```
stevenschlansker: `@Nested` test with an enclosing class run, but don't pick modules from the enclosing test - if this limitation is intentional, consider documenting it, otherwise something like 

```java
InjectTest injectTest = null;
for (Class<?> c = testClass; c != null && injectTest == null; c = c.getEnclosingClass()) {
    injectTest = c.getAnnotation(InjectTest.class);
}
```

should make it work. If so, `profiles` might need a similar update to maintain consistency.

inject-test/src/main/java/io/avaje/inject/test/MetaInfo.java
```diff
@@ -4,16 +4,25 @@
...
+    // read from the test class rather than the instance so that the selected modules also
+    // bound the class level scope of a test that only has static injection
+    this.modules = injectTest == null ? NO_MODULES : injectTest.modules();
```
stevenschlansker, line 26: Agree on this point, but it leaves `profiles` and `scopePerTest` reading from the test instance. For consistency, consider having these two also pull from the annotation and skip reading it off the testInstance.

blackbox-multi-scope/src/test/java/org/multi/main/BoundedCustomModuleTest.java
```diff
@@ -0,0 +1,26 @@
...
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+/**
+ * Bound the wiring to a custom scope module. Custom scope modules are not registered as an {@code
+ * InjectExtension} service, so this covers the reflection fallback in {@code SelectedModules}.
+ */
+@InjectTest(modules = Mod1Module.class)
+class BoundedCustomModuleTest {
+
+  static int constructedBefore;
```
stevenschlansker: this field looks dead?

stevenschlansker (commented):
Thanks for looking at this issue!

PR #1064, opened 2026-08-19
blackbox-conditional-provider-order: repro for conditional in-module provider suppressing the requires edge

stevenschlansker:
A conditional bean that is the only in-module provider of a type consumed by another bean in the same module satisfies that dependency at compile time: the generated module emits no requiresBeans/softRequiresBeans entry for the type while providesBeans claims it unconditionally. When the condition is false at runtime and the real provider lives in another module, no ordering edge exists and wiring fails ("Injecting null") whenever the consuming module wires first, under explicit module lists and classpath discovery alike. Reproduces on 12.6, 12.7-RC2, and the delay-soft ordering (PR #1061).

A consumer in a separate module with a declared @InjectModule(requires) orders correctly past multiple conditional claimants, so the defect is the suppressed requirement, not conditional multi-provider resolution.


stevenschlansker:
#1065  carries the same changes as this MR, so if #1061 merges, this MR is redundant and may be closed.

PR #1067, opened 2026-08-19
Jakarta2Javax: include blackbox-conditional-provider-order child poms

stevenschlansker:
536f8590 added only the aggregator pom to the rewrite lists, so the child poms kept their literal jakarta-flavor parent version through the javax pass and m1 failed to compile against the rewritten javax.inject imports (the build-21 CI leg's 'package javax.inject does not exist'). List all five poms, mirroring blackbox-collection-module-order, in both Jakarta2Javax and Javax2Jakarta. Verified locally: javax-pass reactor build green including the module's four tests, and the reverse pass leaves a clean tree.


SentryMan:
Yeah I dropped the whole hardcoding thing in my latest commit

stevenschlansker:
Awesome, thanks!

